PR #9, opened 2019-10-08
[verilog] Relax ternary parentheses

vogelpi:
This PR relaxes the requirement for parentheses in nested ternary expressions. They should only be recommended but not mandatory. The PR also adds an example for when it is acceptable to avoid the parentheses.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -751,17 +751,14 @@ operations unambiguous.
 
 #### Ternary Expressions
 
-***Nested ternary expressions must be enclosed in parentheses.***
+***Ternary expressions nested in the true condition of another ternary
+expression must be enclosed in parentheses.***
 
 For example:
 
 &#x1f44d;
 ```systemverilog {.good}
-assign state_next = condition_b ?
-    (condition_a ?
-        a_and_b   :
-        b_and_not_a) :
-    state;
+assign foo = condition_a ? (condition_a_x ? x : y) : b;
 ```
 
 While the following nested ternary has only one meaning to the compiler, the
@@ -769,11 +766,16 @@ meaning can be unclear and error-prone to humans:
 
 &#x1f44e;
 ```systemverilog {.bad}
-assign state_next = condition_b ?
-    condition_a ?
-    a_and_b :
-    b_and_not_a :
-    state;
+assign foo = condition_a ? condition_a_x ? x : y : b;
+```
+
+***Parentheses may be omitted if the code formatting conveys the same
+information, for example when describing a priority mux.***
+
+&#x1f44d;
+```systemverilog {.good}
+assign foo = condition_a ? a :
+             condition_b ? b : not_a_nor_b;
 ```
 
 ### Comments


VerilogCodingStyle.md
```diff
@@ -776,6 +776,15 @@ assign state_next = condition_b ?
     state;
 ```
 
+However, the nested ternary can be written in an unambiguous way in many
+cases. Parenthesis are then not required.
+
+&#x1f44d;
+```systemverilog {.good}
+assign a_or_b = condition_a ? a :
+                condition_b ? b : not_a_nor_b;
```
imphil: Possibly indent the second line here?

VerilogCodingStyle.md
```diff
@@ -776,6 +776,15 @@ assign state_next = condition_b ?
     state;
 ```
 
+However, the nested ternary can be written in an unambiguous way in many
+cases. Parenthesis are then not required.
```
imphil: Maybe the other way around, e.g. something like `Parentheses may be omitted if the code formatting conveys the same information.`

And nit: one sentence per line.

asb: I like that rewording.

VerilogCodingStyle.md
vogelpi: As just discussed: This is what we mostly use in our code base. I am thus leaving it.

VerilogCodingStyle.md
vogelpi: Yep. Done.

vogelpi:
Thanks for your feedback @asb and @imphil .

VerilogCodingStyle.md
GregAC: Personally I feel keeping the conditions lined up keeps things more readable

VerilogCodingStyle.md
```diff
@@ -751,7 +751,7 @@ operations unambiguous.
 
 #### Ternary Expressions
 
-***Nested ternary expressions must be enclosed in parentheses.***
+***It is recommended to enclose nested ternary expressions in parentheses.***
```
GregAC: Do we actually want to recommend it? I normally find that with nested ternary's you most often use the style with parentheses omitted below where you're building a priority mux. The ones where you do want parantheses are the rarer ones where the true condition of the ternary has a ternary itself:


```
assign foo = condition_a ? (condition_a_x ? x : y) : b;
```

Or blending the two:

```
assign foo = condition_a ? (condition_a_x ? x : y) :
             condition_b ? b                       :
                           not_a_nor_b; 
```

So it's not a case of recommended or not recommended but rather how you are nesting the ternarys that should guide the use of parentheses.

vogelpi:
@imphil , @asb, would you be happy with the current version? If yes, I will fixup the previous commits with the last one and merge.

asb:
Let's gate this on feedback from @sjgitty

vogelpi:
Thanks for all your feedback guys. I am merging this now.

PR #11, opened 2019-10-24
[verilog] Extend section on comments with header-style

msfschaffner:
This PR adds guidance on how header-style comments should look like in SystemVerilog.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -797,6 +797,71 @@ endmodule : foo
 localparam bit ValBaz = 1;  // This comment describes the item to the left.
 ```
 
+It can sometimes be useful to structure the code using header-style comments in
+order to separate different functional parts (like FSMs, the main datapath or
+registers) within a module. In that case, the preferred style is a single-line
+section name, framed with `//` C++ style comments as follows:
+
+```systemverilog
+module foo;
+
+  ////////////////
+  // Controller //
+  ////////////////
+  ...
+
+  ///////////////////////
+  // Main ALU Datapath //
+  ///////////////////////
+  ...
+
+endmodule : foo
+```
+
+If the designer would like to use comments to mark the beginning/end of a
+particular section for better readability (e.g. in nested for loop blocks), the
+preferred way is to use a single-line comment with no extra delineators, as
+shown in the examples below.
+
+&#x1f44d;
+```systemverilog {.good}
+// begin: iterate over foobar
+for (...) begin
+...
+end
+// end: iterate over foobar
+```
+
+&#x1f44d;
+```systemverilog {.good}
+for (...) begin // iterate over foobar
+...
+end // iterate over foobar
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+//-------------------------- iterate over foobar -------------------------------
+for (...) begin
+...
+end
+//-------------------------- iterate over foobar -------------------------------
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+///////////////////////////////
+// begin iterate over foobar //
+///////////////////////////////
+for (...) begin
+...
+end
+///////////////////////////////
+// end iterate over foobar   //
+///////////////////////////////
+```
+
+
 ### Declarations
 
 ***Signals must be declared before they are used. This means that


VerilogCodingStyle.md
```diff
@@ -797,6 +797,36 @@ endmodule : foo
...
+```
+
+If the designer would like to use comments to mark the beginning/end of a
+particular section for better readability (e.g. in nested for loop blocks), the
+preferred way is to use a single-line comment:
+
+```
+// begin: section
+...
+// end: section
```
imphil: I'm not sure I fully understand that one. Is "section" meant to be replaced by something, or the whole thing, or is section a keyword?

msfschaffner: yes this can be replaced freely, should I change to `<my section id>` or similar ?

imphil: so people are expected to use `begin: my_fsm` or `begin: a awesome code block`? (so an ID, or a description?) Maybe just use an example instead of "section".

msfschaffner: they are not expected to do that on every block, only if they feel that it helps to read the code.

in case they want to add such comments, however, they should use single-line comments similar to the ones shown, without other fancy delineators inside the comment.

let me rephrase this.

msfschaffner:
Ok I added some examples. Let me know if you have other suggestions or feel that one of them is too restrictive.

martin-lueker:
LGTM.   (Is it poor form to volunteer a review?)

asb:
> LGTM. (Is it poor form to volunteer a review?)

Not at all, it's much appreciated.

VerilogCodingStyle.md
```diff
@@ -797,6 +797,70 @@ endmodule : foo
...
+It can sometimes be useful to structure the code using header-style comments in
+order to separate different functional parts (like FSMs, the main datapath or
+registers) within a module. In that case, the preferred style is as follows:
+
+```systemverilog
+module foo;
+
+  ////////////////
+  // Controller //
+  ////////////////
```
sjgitty: I think personally it is OK to also "standardize" on more simple forms of this, such as

```
//////////////
// Controller
```
If people feel like they have to follow this style to the letter, they'll hate the
need to count their slashes, even with guidance from their editor. MO

martin-lueker: Hi Scott.   I agree with hating to count anything, but isn't the original point of this all to squelch individuality in comments?  If we must have such guidelines, I personally would prefer them to be clear and easy to understand.   Perhaps make the style guide clear, and then as an organization discourage excessive nitpicky-ness in enforcement?

msfschaffner: I agree, let me add a couple of permissible variations.

msfschaffner:
Ok, I added a few more examples that are permissible. PTAL...

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+  ///////////////////////////////////////////////////////
+  ...
+
+  ///////////////////////////////////////////////////
+  // It is ok to omit the slashes on the right side
+  ///////////////////////////////////////////////////
+  ...
+
+  case (...)
+    /////////////////////////////////////////
```
eunchan: I feel this should be just a comment. Use of subheadings should be restrictive and not over used.

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+```systemverilog
+//-------------------------- iterate over foobar -------------------------------
+for (...) begin
+...
+end
+//-------------------------- iterate over foobar -------------------------------
+```
+
+&#x1f44e;
+```systemverilog
```
eunchan: Let's not recommend this nor permitted.

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+  ///////////////////////
+  ...
+
+  ///////////////////////////////////////////////////////
+  // Complicated part of the design that warrants some //
+  // more description in the header comment            //
+  ///////////////////////////////////////////////////////
+  ...
+
+  ///////////////////////////////////////////////////
```
eunchan: Will we allow the mismatch of length in this line and the actual heading sentence?

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+  // Controller //
+  ////////////////
+  ...
+
+  ///////////////////////
+  // Main ALU Datapath //
+  ///////////////////////
+  ...
+
+  ///////////////////////////////////////////////////////
```
eunchan: If description is needed, it should be a separate comment below the subheading in my opinion.

VerilogCodingStyle.md
msfschaffner: I would say yes because then you can play with the separator width, even if the text inside the heading is super short.

VerilogCodingStyle.md
msfschaffner: other opinions? I do not feel super strong about this, but have used it in the past...

VerilogCodingStyle.md
msfschaffner: it is actually not permitted (the `&#x1f44e;` is a thumbs down symbol)

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+```
+
+&#x1f44d;
+```systemverilog
+for (...) begin // iterate over foobar
+...
+end // iterate over foobar
+```
+
+&#x1f44e;
```
msfschaffner: this one is also an example which would not be permitted under this style guide...

VerilogCodingStyle.md
msfschaffner: other opinions? I am fine both with keeping or removing it..

eunchan:
```
//////////////////
// Main Heading //
//////////////////
// Description: blah blah
//     extended descriptions.
```

VerilogCodingStyle.md
eunchan: So.. For instance, this is allowed.

```
////////////////////////////////////////////
// It is ok to omit slashes
////////////////////////////////////////////
```

VerilogCodingStyle.md
eunchan: I felt it is enough to

```
  case (...)
    // This is ok as well, e.g. in order to
    // describe what an FSM state does in a
    // switch-case statement
    SomeState: begin
```

VerilogCodingStyle.md
eunchan: Oh... there's slight differences between below and above.. `1f44d` and `1f44`**e**

VerilogCodingStyle.md
msfschaffner: yes this would be allowed

VerilogCodingStyle.md
msfschaffner: sure, we can do that.

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+endmodule : foo
+```
+
+If the designer would like to use comments to mark the beginning/end of a
+particular section for better readability (e.g. in nested for loop blocks), the
+preferred way is to use a single-line comment with no extra delineators, as
+shown in the examples below.
+
+&#x1f44d;
+```systemverilog
```
sjgitty: @msfschaffner can you append ` {.good}` to this? it renders green on some viewers (though not yet on github). See other sections of this document for example

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+// begin: iterate over foobar
+for (...) begin
+...
+end
+// end: iterate over foobar
+```
+
+&#x1f44d;
+```systemverilog
```
sjgitty: ditto `{.good}`

VerilogCodingStyle.md
```diff
@@ -797,6 +797,92 @@ endmodule : foo
...
+for (...) begin // iterate over foobar
+...
+end // iterate over foobar
+```
+
+&#x1f44e;
+```systemverilog
```
sjgitty: ditto except `{.bad}`

VerilogCodingStyle.md
sjgitty: ditto `{.bad}`

VerilogCodingStyle.md
sjgitty: Actually, I think it is fine to keep this since it does say "header-style comments"
in "different functional parts". Not just any standard comment. Fine to standardize
here as much as possible.

VerilogCodingStyle.md
sjgitty: yeah, this is aimed at a particular case: section breaks; not the general case
of commenting code.

msfschaffner: ok I'll remove this.

VerilogCodingStyle.md
msfschaffner: ok will remove it

VerilogCodingStyle.md
msfschaffner: Ok I remove this as well to standardize it as much as possible.

VerilogCodingStyle.md
msfschaffner: jep ;)

msfschaffner:
Ok, another round up for review.

VerilogCodingStyle.md
vogelpi: I am also with Martin here. I would prefer to agree on one single style.

VerilogCodingStyle.md
```diff
@@ -797,6 +797,76 @@ endmodule : foo
...
+  ////////////////
+  ...
+
+  ///////////////////////
+  // Main ALU Datapath //
+  ///////////////////////
+  ...
+
+  //////////////////////////////////////////////////////////////////////////////
+  //        It is ok to play with the width of the comment like this          //
```
asb: I don't see any advantage in allowing this variant. It doesn't save people work, it reduces uniformity, and it just opens up the possibility of the text being misaligned etc.

msfschaffner: ok good I restrict it to one single style.

msfschaffner:
Thanks for your comments, seems like we converged on this one. I continue and merge this in now.

PR #12, opened 2019-12-06
[X/case] This refines guidance on usage of X and case statements

msfschaffner:
Based on our internal discussion this week, this PR refines the guidance on the usage of `unique case` statements and `X` literals in RTL. 

In particular, the guidance on `X` has been completely revised in order to favor a style that eliminates `X` literals in RTL as much as possible, and encourages the extensive use of SVAs to check invalid behavior in RTL instead. 

Note that this guidance shall not affect DV code that is not meant to be synthesized, and `X` can still be used if it is safe to do so.

Signed-off-by: Michael Schaffner <msf@google.com>

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -1784,7 +1784,7 @@ always_comb begin
       end
     end
     StResult: state_d = Idle;
-    default:  state_d = 'X;
+    default: ;
   endcase
 end
 ```
@@ -1797,92 +1797,153 @@ increment.
 
 ### Don't Cares (`X`'s)
 
-***Explicitly specify don't cares when safe to do so. Don't silently
-squash `X` in logic.***
+***The use of `X` literals in RTL code is strongly discouraged. RTL must not
+assert `X` to indicate "don't care" to synthesis in any case.  In order to flag
+and detect invalid conditions, rather than assign and propagate `X` values,
+designs should fully define all signal values and make extensive use of SVAs to
+indicate the invalid conditions. ***
 
-Don't Care values can significantly help the quality of logic optimization and
-should be used wherever it is safe to do so. Explicitly declaring unused decodes
-as `X` has the secondary effect of propagating `X`'s through the design if the
-**input unused** design assumption is violated, making those bugs easier to
-diagnose and fix.
+If not strictly controlled, the use of `X` assignments in RTL to flag invalid or
+don't care conditions can lead to simulation/synthesis mismatches.
 
-Example:
+Instead of assigning and propagating `X` in order to flag and detect invalid
+conditions, it is encouraged to make **extensive use of SVAs**. The added
+benefits of this design practice are that:
+
+- No special code style is required to properly propagate `X` conditions,
+- The chance of accidentally introducing simulation/synthesis mismatches is
+  systematically reduced,
+- Simulation fails quickly and less signal backtracking is needed to root-cause
+  bugs,
+- In several cases, formal property verification (FPV) can be used to prove
+  whether these SVAs can always be fulfilled,
+- In a security context, deterministic/defined behavior is desired, even for
+  illegal/invalid/unreachable input combinations (sometimes stated more tersely
+  as "for security-critical designs, there are no don't-cares").
+
+The solution presented here has similarities with the approaches presented in
+["Being Assertive With Your X"](http://www.lcdm-eng.com/papers/snug04_assertiveX.pdf)
+by Don Mills.
+
+Note that although don't cares can be used to indicate possible optimization
+opportunities to the synthesis tool, it is debatable whether the gains in logic
+reduction are significant enough to outweigh the possible synthesis mismatch
+issues that the use of `X` literals may entail (especially with the gate-counts
+available in today's technologies).
+
+#### Specific Guidance on Case Statements and Ternaries
+
+To comply with this style, RTL must place `` `ASSERT_KNOWN`` assertions on all
+module outputs, with the exception of signals that may implicitly be `X` at the
+beginning of the simulation, such as FIFO, SRAM or register file outputs.
 
 ```systemverilog
+module mymod (
+  input        ina_i,
+  input        inb_i,
+  output logic out_o
+);
+  assign out_o = ina_i ^ inb_i;
+  `ASSERT_KNOWN(OutKnown_A, out_o, clk_i, !rst_ni)
+endmodule : mymod
+```
+
+Further, it is encouraged to add assertions to the signals forming conditions of
+case statements, ternaries or if/else statements. The assertion style is at the
+designer's discretion, and can range from simple `` `ASSERT_KNOWN``  to fully
+functional assertions, as shown in the following examples:
+
+```systemverilog
+typedef enum logic [1:0] {mode0, mode1, mode2} state_e;
+state_e sel;
+
+// encouraged
+`ASSERT_KNOWN(SelKnown_A, sel)
 always_comb begin
-  unique case (state)
-    ALPHA:   decode = 16'd1;
-    BETA:    decode = 16'd127;
-    GAMMA:   decode = 16'd43;
-    // all other states are unused:
-    default: decode = 'X;  // or {16{1'bx}} in Verilog-2001
+  out0 = '0;
+  out1 = '0;
+  unique case (sel)
+    mode1: out0 = foo;
+    mode2: out1 = bar;
+    default: ;
+  endcase
+end
+
+// optional, but more explicit
+// not always applicable
+`ASSERT(MainFsmCase_A, sel inside {mode0, mode1, mode2}, clk_i, !rst_ni)
+always_comb begin
+  out0 = '0;
+  out1 = '0;
+  unique case (sel)
+    mode1: out0 = foo;
+    mode2: out1 = bar;
+    default: ;
   endcase
 end
 ```
 
-Write logic that will either propagate `X` or assert when your inputs are `X`.
-Avoid silently squashing `X` by writing logic that resolves an `X` input to a
-non-`X` output. Instead, write logic that either explicitly propagates the `X`
-or uses an assert to raise an exception when an input is `X`.
-
-Be aware: Any logical operation involving `X` always propagates the `X`.
-
-Example:
+In the context of ternary statements, the following are encouraged examples:
 
 ```systemverilog
-$display("%b", 1'bx == 1'b1);  // produces 1'bx
-$display("%b", 1'bx != 1'b1);  // produces 1'bx
-$display("%b", !(1'bx));       // produces 1'bx.
-// etc.
+// encouraged
+`ASSERT_KNOWN(ModeKnown_A, mode_i, clk_i, !rst_ni)
+`ASSERT_KNOWN(LenKnown_A, len_i, clk_i, !rst_ni)
+// assign '0 for all other combinations
+assign val = (mode_i == ENC)                    ? 8'h01 :
+             (mode_i == DEC && len_i == LEN128) ? 8'h36 :
+             (mode_i == DEC && len_i == LEN192) ? 8'h80 :
+             (mode_i == DEC && len_i == LEN256) ? 8'h40 : 8'h00;
+
+// optional, but more explicit
+`ASSERT(ValSelValid_A, mode_i == ENC || mode_i == DEC &&
+    len_i inside {LEN128, LEN192, LEN256}, clk_i, !rst_ni)
+// using one of the valid outputs for other combinations (saves logic)
+assign val = (mode_i == ENC)                    ? 8'h01 :
+             (mode_i == DEC && len_i == LEN128) ? 8'h36 :
+             (mode_i == DEC && len_i == LEN192) ? 8'h80 :
+             (mode_i == DEC && len_i == LEN256) ? 8'h40 : 8'h01;
 ```
 
-However, when evaluated in a boolean context, `X` always evaluates to false.
+#### Dynamic Array Indexing
 
-Example:
+It should be noted that dynamic array indexing operations can implicitly lead to
+`X`. This should be avoided if possible by either aligning indexed arrays to
+powers of 2 or by adding guarding if statements around the indexing operation.
+These solutions are illustrated in the following examples.
 
-```systemverilog
-always_comb begin
-  if (value) result = 1'b0;    // value is 1'b1
-  else result = 1'b1;          // value is 1'b0 or 1'bx
-end
+
+&#x1f44e;
+```systemverilog {.bad}
+logic selected;
+logic [3:0] idx;
+logic [11:0] foo; // problematic
+
+assign foo = {12'b1010_1111_0000};
+assign selected = foo[idx];
 ```
 
-This can mask subtle problems in code and produce a mismatch between simulation
-and synthesis.
+&#x1f44d;
+```systemverilog {.good}
+logic selected;
+logic [3:0] idx;
+logic [15:0] foo; // aligned to powers of two
 
-Instead, consider these options:
-
-```systemverilog
-always_comb begin
-  result = value ? 1'b1 : 1'b0;  // 1'bx produces 1'bx
-end
+assign foo = {4'b0000, 12'b1010_1111_0000};
+assign selected = foo[idx];
 ```
 
-or
+&#x1f44d;
+```systemverilog {.good}
+logic selected;
+logic [3:0] idx;
+logic [11:0] foo;
 
-```systemverilog
-always_comb begin
-  if (value) result = 1'b0;        // value is 1'b1
-  else if (!value) result = 1'b1;  // value is 1'b0
-  else result = 1'bx;              // value is 1'bx
-end
+assign foo = {12'b1010_1111_0000};
+
+// guarding if statement
+assign selected = (idx < $bits(foo)) ? foo[idx] : 1'b0;
 ```
-
-or
-
-```systemverilog
-always_comb begin
-  assert (!$isunknown(value));     // throws exception if 1'bx
-  if (value) result = 1'b0;        // value is 1'b1
-  else result = 1'b1;              // value is 1'b0 or 1'bx
-end
-```
-
-Further discussion:
-
--   ["I'm Still In Love With My X!"](http://www.sutherland-hdl.com/papers/2013-DVCon_In-love-with-my-X_paper.pdf)
-    by Stuart Sutherland
--   ["Being Assertive With Your X"](http://www.lcdm-eng.com/papers/snug04_assertiveX.pdf) by Don Mills
 
 ### Combinational Logic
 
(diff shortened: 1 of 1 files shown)


msfschaffner:
Note that this is a WIP PR. Once the initial draft has been stabilized, I will invite more reviewers to comment on the changes.

msfschaffner:
Should we mention somewhere in the style guide that `X` are still ok in DV related code that is not meant to be synthesized? I believe this could also solve some of @GregAC's concerns regarding `X` assignments used in bus models.

sjgitty:
> Should we mention somewhere in the style guide that `X` are still ok in DV related code that is not meant to be synthesized? I believe this could also solve some of @GregAC's concerns regarding `X` assignments used in bus models.

@sriyerg is considering (probably has already started) a style guide for verification code. We could put a brief mention here and save the rest for that section / document.

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
+reduction are significant enough to outweigh the possible synthesis mismatch
+issues that the use of `X` literals may entail (especially with the gate-counts
+available in today's technologies). Large LUTs may be an exception to this
+statement and the use of `X` in that context should be carefully evaluated on a
+case-by-case basis.
 
-Example:
+#### Specific Guidance on Case Statements and Ternaries
+
+To comply with this style, RTL must place `` `ASSERT_KNOWN`` assertions on all
```
GregAC: I presume ASSERT_KNOWN is an assertion that a signal isn't X? Have we actually got this assertion defined anywhere? Even if we do we should define what it or point towards a definition in the style guide.

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
+// alternative using inline assertion
+// not always applicable
 always_comb begin
-  result = value ? 1'b1 : 1'b0;  // 1'bx produces 1'bx
+  out0 = '0;
+  out1 = '0;
+  unique case (sel)
+    mode1: out0 = foo;
+    mode2: out1 = bar;
+    default: `ASSERT_I(Invalid_A, 0);
```
GregAC: Need to define what ASSERT_I is

VerilogCodingStyle.md
```diff
@@ -1928,26 +2007,63 @@ always_comb begin
...
 
-The `default` case is required to avoid accidental inference of latches, even if
-all cases are covered. In simulation, a case expression that evaluates to `X`
-will not match any case and will behave as a latch, leading to different
-behavior than synthesis. Instead, in most scenarios, the best choice is to
-propagate the `X` by assigning all output variables to `X` by default.
+  - if no default assignments are given before the case statement as shown in
+  the example above, any variables assigned in one case item must be assigned in
+  all case items, including the `default:`. Failing to do this can lead to a
+  simulation-synthesis mismatch as described in [Don Mills' paper][yalagp].
```
GregAC: Link missing?

VerilogCodingStyle.md
msfschaffner: Yes, `ASSERT_KNOWN` is essentially what you sometimes refer to as `NEVER_X`. Good catch, I think we only have implicit references to these macros in the style-guide. I.e., if you search for `ASSERT`, you'll find another `ASSERT_INIT` and `ASSERT` reference, but these macros are never explained explicitly. I'll amend this and add a short section that explains these.

VerilogCodingStyle.md
msfschaffner: Yep, see previous comment.

VerilogCodingStyle.md
msfschaffner: Oh yeah that got lost...

VerilogCodingStyle.md
```diff
@@ -1928,26 +2007,63 @@ always_comb begin
...
+    end
+    Work: begin
+      state_d = Wait;
+      outb = in1;
+    end
+    Wait: begin
+      state_d = Idle;
+      outc = in2;
+    end
+    default: ; // always assign a default, even if empty!
```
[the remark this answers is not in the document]
msfschaffner: good point, I'll add.

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
+logic [3:0] idx;
+logic [15:0] foo; // aligned to powers of two
+
+assign foo = {4'b0000, 12'b1010_1111_0000};
+assign selected = foo[idx];
 ```
 
-Further discussion:
+&#x1f44d;
+```systemverilog {.good}logic selected;
```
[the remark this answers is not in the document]
msfschaffner: yes, will fix.

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
-Don't Care values can significantly help the quality of logic optimization and
-should be used wherever it is safe to do so. Explicitly declaring unused decodes
-as `X` has the secondary effect of propagating `X`'s through the design if the
-**input unused** design assumption is violated, making those bugs easier to
-diagnose and fix.
+If not strictly controlled, the use of `X` assignments in RTL to flag invalid or
+don't care conditions can lead to simulation/synthesis mismatches.
 
-Example:
+Instead of assigning and propagating `X` in order to flag and detect invalid
```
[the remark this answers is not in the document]
msfschaffner: Indeed, I'll add that.

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
+  unique case (sel)
+    mode1: out0 = foo;
+    mode2: out1 = bar;
+    default: `ASSERT_I(Invalid_A, 0);
+  endcase
 end
 ```
 
-or
+In the context of ternary statements, the following are encouraged examples
```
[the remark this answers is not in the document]
msfschaffner: fixed

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
-However, when evaluated in a boolean context, `X` always evaluates to false.
-
-Example:
+Further, it is encouraged to add assertions to the signals forming conditions of
+case statements, ternaries or if/else statements. The assertion style is at the
+designer's discretion, and can range from simple `` `ASSERT_KNOWN``  to fully
+functional assertions, as shown in the following examples:
 
 ```systemverilog
+// encouraged
```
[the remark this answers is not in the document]
msfschaffner: added

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
-  endcase
-end
-```
+- No special code style is required to properly propagate `X` conditions,
+- The chance of accidentally introducing simulation/synthesis mismatches is
+  systematically reduced,
+- Simulation breaks early and less signal backtracking is needed to
+  root-cause bugs,
+- In several cases, formal property verification (FPV) can be used to prove
+  whether these SVAs can always be fulfilled.
```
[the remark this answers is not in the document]
msfschaffner: thanks @sjgitty and @cdgori, I added this point to the list.

VerilogCodingStyle.md
msfschaffner: actually this is a cross reference within the document, so that link is correct.

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
-as `X` has the secondary effect of propagating `X`'s through the design if the
-**input unused** design assumption is violated, making those bugs easier to
-diagnose and fix.
+If not strictly controlled, the use of `X` assignments in RTL to flag invalid or
+don't care conditions can lead to simulation/synthesis mismatches.
 
-Example:
+Instead of assigning and propagating `X` in order to flag and detect invalid
+conditions, it is encouraged to make **extensive use of SVAs**. The added
+benefits of this design practice is that
```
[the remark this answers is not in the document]
msfschaffner: thanks, fixed

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
-    GAMMA:   decode = 16'd43;
-    // all other states are unused:
-    default: decode = 'X;  // or {16{1'bx}} in Verilog-2001
-  endcase
-end
-```
+- No special code style is required to properly propagate `X` conditions,
+- The chance of accidentally introducing simulation/synthesis mismatches is
+  systematically reduced,
+- Simulation breaks early and less signal backtracking is needed to
```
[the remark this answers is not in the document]
msfschaffner: fixed

VerilogCodingStyle.md
```diff
@@ -1928,26 +2007,63 @@ always_comb begin
...
-    default: operand = 'X;  // propagate X
+    default: operand = '0; // assign a default
   endcase
 end
 ```
 
 The `unique` prefix is recommended before all case statements, as it creates
 simulation assertions that can catch certain mistakes. In some cases, `priority`
-may be used instead of `unique`, though `if-else-if` structures are a more
+may be used instead of `unique`, though in such cases, `if-else-if` structures
```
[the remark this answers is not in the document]
msfschaffner: I replaced this with cascaded ternary structures

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
+by Don Mills.
 
-Be aware: Any logical operation involving `X` always propagates the `X`.
+Note that although don't cares can be used to indicate possible optimization
+opportunities to the synthesis tool, it is debatable whether the gains in logic
+reduction are significant enough to outweigh the possible synthesis mismatch
+issues that the use of `X` literals may entail (especially with the gate-counts
+available in today's technologies). Large LUTs may be an exception to this
+statement and the use of `X` in that context should be carefully evaluated on a
+case-by-case basis.
```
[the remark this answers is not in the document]
msfschaffner: Yeah I completely agree with your point, although I think that should probably go into our additional RTL coding guide for security critical applications.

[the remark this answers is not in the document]
msfschaffner: WDYT, @sjgitty ?

VerilogCodingStyle.md
```diff
@@ -1928,26 +2003,65 @@ always_comb begin
...
+  inference of latches, even if all cases are covered. In simulation, a case
+  expression that evaluates to `X` will not match any case and will behave as a
+  latch, leading to different behavior than synthesis if no default is specified.
+
+  - if no default assignments are given before the case statement as shown in
+  the example above, any variables assigned in one case item must be assigned in
+  all case items, including the `default:`. Failing to do this can lead to a
+  simulation-synthesis mismatch as described in [Don Mills' paper][yalagp].
+
+The following is a different example showing a style-compliant case statement
```
msfschaffner: @sjgitty shall I omit this second case statement example? looks like this is not strictly needed after stating the points above...

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,171 @@ increment.
...
+assign foo = {12'b1010_1111_0000};
+
+// guarding if statement
+assign selected = (idx < $bits(foo)) ? foo[idx] : 1'b0;
+```
 
--   ["I'm Still In Love With My X!"](http://www.sutherland-hdl.com/papers/2013-DVCon_In-love-with-my-X_paper.pdf)
-    by Stuart Sutherland
--   ["Being Assertive With Your X"](http://www.lcdm-eng.com/papers/snug04_assertiveX.pdf) by Don Mills
+#### A Note on Security Critical Applications
```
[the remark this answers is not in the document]
msfschaffner: do you think we should remove this note concerning the assertion macros?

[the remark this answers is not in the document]
msfschaffner: ah and II have added the note you and @cdgori suggested in the section on X-avoidance.

msfschaffner:
Added a revised version of the style guide. 

@mwbranstad, I also added an example that shows which operator to use within the context of flip-flop resets `if (!rst_ni)` as opposed to `if (~rst_ni)`.

imphil:
Can you do the second commit in a separate PR? That one should be easy to agree on and get merged, and will keep the discussion here focused.

VerilogCodingStyle.md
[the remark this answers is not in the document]
GregAC: If by large LUTs you mean large roms encoded with case statements or similar I don't think it's unreasonable to say you must use 0 or 1 for the bits you don't care about (designers choice, e.g. if you want to output 8'b1xxxxxxx then 1 may be the sensible choice as it matches the other bit or if there's lot of inputs that are total don't cares on output then try to match the most used bit say you output either one or two 1s with 0s and otherwise Xs the don't care cases would become zeros).

You could also say that if you end up with a 'sparse LUT' (i.e. one where we don't care about much of the inputs and/or outputs) then you may be doing it wrong and should rethink.

As a final fallback recommend you push the LUT with Xs through some logic minimization tool to get the same LUT back but with choices for the Xs that work well (though I'm not sure such a tool exists you'll just get a bunch of gates and we may get back to the same information leakage issue).

VerilogCodingStyle.md
```diff
@@ -1797,92 +1797,167 @@ increment.
...
+issues that the use of `X` literals may entail (especially with the gate-counts
+available in today's technologies). Large LUTs may be an exception to this
+statement and the use of `X` in that context should be carefully evaluated on a
+case-by-case basis.
+
+#### Specific Guidance on Case Statements and Ternaries
+
+To comply with this style, RTL must place `` `ASSERT_KNOWN`` assertions on all
+module outputs, with the exception of signals that may implicitly be `X` at the
+beginning of the simulation, such as FIFO, SRAM or register file outputs.
```
GregAC: In the separate security guidelines we way want to say all such structures should receive an explicit reset to 0 via FSM during an init phase?

VerilogCodingStyle.md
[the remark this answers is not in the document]
msfschaffner: Indeed, I meant variations of large case statements or arrays.

In the past, people have been using tools like [Espresso](https://en.wikipedia.org/wiki/Espresso_heuristic_logic_minimizer) to minimize large truth tables - although in my own experiments I got the best results by specifying large LUTs with arrays in SystemVerilog and running them straight through DC. 

I never really specified sparse LUTs in this manner, but I did use truncated arrays for unaligned LUTs. I.e., say you have an LUT with 567 words, I would use an array `[0:566]`, which implicitly means that `[567:1023]` are `X` when indexed with higher indexes. But then again padding with zero up to the next pow2 depth, or using a guarding if statement around the indexing operation would likely not add much more logic and would make this predictable and safe.

So I would say, let's make this strict and just remove that statement on LUTs.

VerilogCodingStyle.md
msfschaffner: That is another question for @cdgori :)...

msfschaffner:
Removed the sentence stating that large LUTs may be an exception to the new guidance on `X`.

msfschaffner:
Removed the following recommended example, as it uses an unclocked assertion which may be susceptible to glitches in simulation:
```systemverilog
// alternative using inline assertion
always_comb begin
  out0 = '0;
  out1 = '0;
  unique case (sel)
    mode1: out0 = foo;
    mode2: out1 = bar;
    default: `ASSERT_I(Invalid_A, 0);
  endcase
end
```
See also https://github.com/lowRISC/ibex/pull/528 for context.

VerilogCodingStyle.md
```diff
@@ -1928,26 +1989,65 @@ always_comb begin
...
-propagate the `X` by assigning all output variables to `X` by default.
+```systemverilog
+always_comb begin
+  // common default assignments
+  state_d = state_q;
+  outa = 1'b0;
+  outb = 1'b0;
+  outc = 1'b0;
+
+  unique case  (state_q)
```
sjgitty: nit: double-space here

msfschaffner: thanks, fixed.

msfschaffner:
I am putting this up for review after this initial round of comments. PTAL.

We are likely going to do a general editing pass in the future to make the style guide more compact - but this will be a separate PR.

msfschaffner:
Squashed.

msfschaffner:
Are there any other comments that should be addressed before moving this one in?

@GregAC can you give this another look and unblock the PR if you are happy with the current state?

VerilogCodingStyle.md
sjgitty: I think it is fine as written now, thanks.

VerilogCodingStyle.md
```diff
@@ -1928,26 +1989,65 @@ always_comb begin
...
+  latch, leading to different behavior than synthesis if no default is specified.
+
+  - if no default assignments are given before the case statement as shown in
+  the example above, any variables assigned in one case item must be assigned in
+  all case items, including the `default:`. Failing to do this can lead to a
+  simulation-synthesis mismatch as described in [Don Mills' paper][yalagp].
+
+The following is a different example showing a style-compliant case statement
+variant that is frequently used for describing the next-state logic of a finite
+sate machine. What is different from the previous example is that the default
```
sjgitty: typo: ^sate^state^

VerilogCodingStyle.md
```diff
@@ -2406,6 +2518,45 @@ always_ff @(posedge clk) begin
...
+// (i.e. that the signal is not `X`).
+`ASSERT_KNOWN(<name>, <signal>, <clk>, <reset condition>)
+```
+
+An implementation of these macros (including other useful variations thereof)
+can be found here:
+https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_assert.sv
+
+
+#### A Note on Security Critical Applications
```
sjgitty: Personally I think we can skip this section until we have full guidance here. But let's go ahead and leave it here, but please add something like "More security assertion and coding style guidance will be coming" else this seems like an odd incomplete story on the topic.

VerilogCodingStyle.md
msfschaffner: fixed, thanks

VerilogCodingStyle.md
msfschaffner: Ok I added the following sentence at the end of this subsection:
```
More security assertion and coding style guidance will be given in a separate
document, soon.
```

msfschaffner:
> A few comments, but otherwise this is looking good to me. If there isn't an issue already filed, please file one to create security-related coding style and assertion guidance, CCing the usual suspects.

Thanks for giving it another look, I've filed an issue.

sriyerg:
> > Should we mention somewhere in the style guide that `X` are still ok in DV related code that is not meant to be synthesized? I believe this could also solve some of @GregAC's concerns regarding `X` assignments used in bus models.
> 
> @sriyerg is considering (probably has already started) a style guide for verification code. We could put a brief mention here and save the rest for that section / document.

@udinator, let's ensure that we address this in the DV style guide.

msfschaffner:
Ok it looks like we converged on this one - merging.

Issue #19, opened 2020-01-21
Remove restriction of wildcard `import my_pkg::*;` inside the same module


eunchan:
I would like to remove the clause that restricts the usage of wildcard import of the package.

In some cases, it is quite useful not to type the package namespace every time.
The concerns causing the restriction above are the collision of the names between the module and the package and the usage of wildcard import outside of the module. But if the wildcard import is used inside the module and the same IP that defines the package, it is unlikely to have collisions.

So I would like to suggest below cases are permitted in the style guide.

1. wildcard import at the module header is permitted if the package is defined in the same IP.

```systemverilog

module mod_a import mod_a_pkg::*; (
);

endmodule
```
2. wildcard import is permitted inside the module body if the package is defined in the same IP.

```systemverilog
module mod_a ();

  import mod_a_pkg::*;

endmodule
```

Other than the cases above, wildcard import is not allowed. For instance, below example can create name collision in the module following mod_a in the source list.

```systemverilog
// mod_a.sv
import mod_a_pkg::*;

module mod_a ();


// mod_b.sv
module mod_b ();
  // mod_b cannot use the name defined in mod_a_pkg
endmodule
```

Other bad examples:

```systemverilog
// wildcard import for other packages outside of the IP
module mod_a import mod_b_pkg::*; ();
```

```systemverilog
module mod_a ();

  // wildcard import of different packages.
  import mod_b_pkg::*;

endmodule
```

**directory structure for above examples**

```console
hw/ip
    | mod_b/rtl/mod_b_pkg.sv
    \ mod_a/rtl/
        | mod_a.sv
        \ mod_a_pkg.sv
```

CC: @tjaychen @msfschaffner

msfschaffner:
Indeed, I support this proposal. This issue has in part also been brought up by @mwbranstad (we just did not yet file it as an issue on GH).

mwbranstad:
Agreed. My original comment to @msfschaffner was:
•	Wildcard packages: Wildcard import (of packages), eg. import my_pkg::*;.
o	This style is being commonly used for registers, so this statement is a bit confusing.

GregAC:
We do already have examples of 2. in Ibex so if we're deciding not to allow this we have some cleaning up to do!

I'm in favor of this with the rules @eunchan specifies.

tomroberts-lowrisc:
+1 from me

asb:
I've got no objection either. I think this is good to go to a PR. Will you be following up with one @eunchan?

[...]

PR #20, opened 2020-01-22
Remove wildcard import restriction

eunchan:
Previously, in the Verilog Style Guide, wildcard import is not
recommended. But the designs at lowRISC repo use the wildcard import in
certain conditions.

This PR is to alleviate the restriction in some conditions.

1. Allow wildcard import at the module header if the imported package is
   in the same IP.
2. Allow wildcard import at the body if the imported package is in the
   same IP.

Please remind that any style of import for the package in the same IP is
not allowed if the import syntax is used out of the module.

```systemverilog {.bad}
// bad example
import my_pkg::str_t;

module my_module ();
```

This PR closes issue #19

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -116,6 +116,7 @@ representation of this style guide.
     - [Active-Low Signals](#active-low-signals)
     - [Differential Pairs](#differential-pairs)
     - [Delays](#delays)
+    - [Wildcard import of packages](#wildcard-import-of-packages)
   - [Appendix - Condensed Style Guide](#appendix---condensed-style-guide)
     - [Basic Style Elements](#basic-style-elements)
     - [Construct Naming](#construct-naming)
@@ -2188,7 +2189,6 @@ These language features are considered problematic and their use is discouraged
 unless otherwise noted:
 
 -   Interfaces.
--   Wildcard import (of packages), eg. `import my_pkg::*;`.
 -   The `alias` statement.
 -   `case inside` is broken inside some FPGA compile tools.
 
@@ -2492,6 +2492,85 @@ always_ff @(posedge clk) begin
   data_valid_q2 <= data_valid_q;
   data_valid_q3 <= data_valid_q2;
 end
+```
+
+### Wildcard import of packages
+
+The wildcard import syntax, e.g. `import ip_pkg::*;` is only allowed where the
+package is part of the same IP as the module that uses that package.  Wildcard
+import statement must be placed in the module header or in the module body.
+
+&#x1f44d;
+```systemverilog {.good}
+// mod_a_pkg.sv and mod_a.sv are in the same IP.
+// Packages can be imported in the module declaration if access to
+// unqualified types is needed in the port list.
+
+// mod_a_pkg.sv
+package mod_a_pkg;
+
+  typedef struct packed {
+    ...
+  } a_req_t;
+endpackage
+
+// mod_a.sv
+module mod_a
+  import mod_a_pkg::*;
+(
+  ...
+  a_req_t a_req,
+  ...
+);
+
+endmodule
+```
+
+&#x1f44d;
+```systemverilog {.good}
+// mod_a
+module mod_a ();
+
+  // mod_a_pkg.sv and mod_a.sv are in the same IP.
+  import mod_a_pkg::*;
+
+  ...
+
+  a_req_t a_req;
+
+endmodule
+```
+
+Other than the cases above, wildcard imports are not allowed. For instance,
+the example below may create a name collision in the module following `mod_a`
+in the source list.
+
+&#x1f44e;
+```systemverilog {.bad}
+// mod_a.sv
+import mod_a_pkg::*; // not allowed: imported to $root scope.
+
+module mod_a ();
+
+endmodule
+```
+
+Other bad examples:
+
+&#x1f44e;
+```systemverilog {.bad}
+// wildcard import for other packages outside of the IP
+module mod_a import mod_b_pkg::*; ();
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+module mod_a ();
+
+  // not allowed: wildcard import of a package from a different IP
+  import mod_b_pkg::*;
+
+endmodule
 ```
 
 ### Assertion Macros


VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
...
+### Wildcard import of packages
+
+It is not permitted to use wildcard import syntax, eg. `import other_pkg::*;`
+in general to avoid naming collisions. However, wildcard import is allowed if
+the imported package is placed at the same IP and the import statement is
+placed at the module header or the module body.
+
+&#x1f44d;
+```systemverilog {.good}
+// mod_a_pkg.sv and mod_a.sv are placed at the same directory
```
imphil: Maybe add a comment here what this syntax is good for and when to prefer it. Effectively: "Packages can be imported in the module declaration if access to unqualified types is needed in the port list."

VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
...
+```
+
+Other than the cases above, wildcard import is not allowed. For instance, below
+example may create name collision in the module following mod_a in the source
+list.
+
+&#x1f44e;
+```systemverilog {.bad}
+// mod_a.sv
+import mod_a_pkg::*; // not allowed: affecting all modules following
```
imphil: IIRC the terminology the LRM uses here is `$root` scope.

VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
...
+
+module mod_a ();
+
+
+// mod_b.sv
+module mod_b ();
+  // mod_b cannot use the name defined in mod_a_pkg
+endmodule
```
imphil: I know what you mean, but it's slightly confusing as this is effectively the reasoning, and not necessary for the example any more. I'd remove the `mod_b` part from this example.

VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
...
+```systemverilog {.bad}
+// wildcard import for other packages outside of the IP
+module mod_a import mod_b_pkg::*; ();
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+module mod_a ();
+
+  // wildcard import of different packages.
```
imphil:
```suggestion
  // not allowed: wildcard import of a package from a different directory
```

VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
 end
 ```
 
+### Wildcard import of packages
+
+It is not permitted to use wildcard import syntax, eg. `import other_pkg::*;`
```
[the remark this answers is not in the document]
GregAC: As wildcard import syntax is permitted under some circumstances I think the opening should be more like

'Wildcard import syntax, e.g. 'import ip_pkg::*;' is only allowed where the package is part of the same IP as the module that uses that package. Wildcard import statement must be placed in the module header or module body'

VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
...
+in general to avoid naming collisions. However, wildcard import is allowed if
+the imported package is placed at the same IP and the import statement is
```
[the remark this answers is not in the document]
GregAC: I'd prefer something more like 'IP'. For example if we decided to split up the ibex RTL into a few sub-directories I think we should still allow all of the modules to use ibex_pkg::* imports.

VerilogCodingStyle.md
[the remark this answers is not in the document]
eunchan: Thanks for suggestion! Revised

VerilogCodingStyle.md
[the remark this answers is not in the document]
eunchan: Right. intention is IP as broader term. I will change 'directory' to IP

VerilogCodingStyle.md
```diff
@@ -2494,6 +2493,68 @@ always_ff @(posedge clk) begin
...
+placed at the module header or the module body.
+
+&#x1f44d;
+```systemverilog {.good}
+// mod_a_pkg.sv and mod_a.sv are placed at the same directory
+module mod_a import mod_a_pkg::*; (
```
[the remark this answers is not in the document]
eunchan: Good

VerilogCodingStyle.md
eunchan: OK added comment and revised the example to show the usage case.

VerilogCodingStyle.md
eunchan: Thanks! added `$root` scope here

eunchan:
@imphil Could you please take an another look?

VerilogCodingStyle.md
```diff
@@ -2494,6 +2494,84 @@ always_ff @(posedge clk) begin
 end
 ```
 
+### Wildcard import of packages
+
+Wildcard import syntax, e.g. `import ip_pkg::*;` is only allowed where the
```
imphil:
```suggestion
The wildcard import syntax, e.g. `import ip_pkg::*;` is only allowed where the
```

VerilogCodingStyle.md
```diff
@@ -2494,6 +2494,84 @@ always_ff @(posedge clk) begin
...
+package is a part of the same IP as the module that uses that package. Wildcard
```
imphil:
```suggestion
package is part of the same IP as the module that uses that package. The wildcard
```

VerilogCodingStyle.md
```diff
@@ -2494,6 +2494,84 @@ always_ff @(posedge clk) begin
...
+import statement must be placed in the module header or module body.
```
imphil:
```suggestion
import statement must be placed in the module header or in module body.
```

VerilogCodingStyle.md
```diff
@@ -2494,6 +2494,84 @@ always_ff @(posedge clk) begin
...
+package mod_a_pkg;
+
+  typedef struct packed {
+    ...
+  } a_req_t;
+endpackage
+
+// mod_a.sv
+module mod_a
+  import mod_a_pkg::*; (
```
imphil:
```suggestion
  import mod_a_pkg::*; 
(
```

PR #22, opened 2020-02-06
Initial commit of the Verilog DV style guide

udinator:
Signed-off-by: Udi Jonnalagadda <udij@google.com>

### DIFF
--- a/DVCodingStyle.md
+++ b/DVCodingStyle.md
@@ -1,0 +1,880 @@
+# lowRISC SystemVerilog Coding Style Guide for Design Verification
+
+This document is a supplemental Style Guide to the [lowRISC Verilog style
+guide](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md),
+with emphasis on writing code for Design Verification (DV) in SystemVerilog,
+following the [UVM methodology](https://www.accellera.org/images//downloads/standards/uvm/uvm_users_guide_1.2.pdf).
+
+## Table of Contents
+
+* [Introduction](#introduction)
+* [Naming and Style](#naming-and-style)
+* [Directory Structure and Packages](#directory-structure-and-packages)
+  * [UVM Agent Directory Structure](#uvm_agent-directory-structure)
+  * [UVM Env Directory Structure](#uvm_env-directory-structure)
+* [UVM Guidelines](#uvm-guidelines)
+  * [Class Definitions](#class-definitions)
+  * [UVM new functions](#uvm-new-functions)
+  * [uvm_scoreboard usage](#uvm_scoreboard-usage)
+  * [uvm_agent usage](#uvm_agent-usage)
+  * [uvm_driver usage](#uvm_driver-usage)
+  * [Macro Usage](#macro-usage)
+    * [Guidelines for UVM library macros](#guidelines-for-uvm-library-macros)
+    * [Guidelines for macros within dv_macros.svh](#guidelines-for-macros-within-dv_macrossvh)
+    * [General macro use guidelines](#general-macro-use-guidelines)
+  * [Factory](#factory)
+  * [Configuration Mechanism](#configuration-mechanism)
+  * [Sequences](#sequences)
+  * [Sequence Items](#sequence-items)
+  * [RAL Usage](#ral-usage)
+  * [Objections and Coordinating End of Test](#objections-and-coordinating-end-of-test)
+  * [Logging and Print Messages](#logging-and-print-messages)
+  * [DPI and C Connections](#dpi-and-c-connections)
+* [SystemVerilog Language Features](#systemverilog-language-features)
+  * [Function Declarations](#function-declarations)
+  * [Randomization](#randomization)
+  * [Enums](#enums)
+  * [Interfaces, Clocking Blocks, Modports](#interfaces,-clocking-blocks,-modports)
+  * [Loop Operators](#loop-operators)
+  * [Code Within Asserts](#code-within-asserts)
+  * [Wait and Fork](#wait-and-fork)
+  * [Void Casts](#void-casts)
+  * [Associative Arrays](#associative-arrays)
+  * [Bind Statements](#bind-statements)
+  * [Simulator Specific Code](#simulator-specific-code)
+  * [Forbidden System Tasks and Functions](#forbidden-system-tasks-and-functions)
+
+
+
+## Introduction
+
+This guide defines the Comportable style for DV-specific SystemVerilog code. The
+goal is to use this common set of guidelines when implementing testbenches using
+the UVM methodology to allow all developed Comportable DV code to be uniform,
+reusable, and portable.
+This document assumes previous experience and working knowledge of UVM, and
+as such all included code snippets will be fairly concise.
+
+
+## Naming and Style
+
+Refer to [lowRISC Verilog naming
+conventions](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#naming)
+for any general naming style guidelines. In general, common UVM testbench
+components must be declared and assigned handles as follows:
+
+| Testbench component               | Style                                        |
+| --------------------------------- | -------------------------------------------- |
+| Virtual interfaces                | `virtual <if_name>_if <if_name>_vif;`        |
+| Environment                       | `<dut>_env env;`                             |
+| Env level config object           | `rand <dut>_env_cfg cfg;`                    |
+| Env coverage collection object    | `<dut>_env_cov cov;`                         |
+| Agent                             | `<dut>_agent m_<dut>_agent;`                 |
+| Agent level config object         | `rand <dut>_agent_cfg m_<dut>_agent_cfg;`    |
+| Agent coverage collection object  | `<dut>_agent_cov cov;`                       |
+| Driver                            | `<dut>_driver driver;`                       |
+| Monitor                           | `<dut>_monitor monitor;`                     |
+| Scoreboard                        | `<dut>_scoreboard scoreboard;`               |
+| Virtual sequencer                 | `<dut>_virtual_sequencer virtual_sequencer;` |
+| Sequencer                         | `<dut>_sequencer <dut>_sequencer;`           |
+
+Additional naming guidelines:
+
+1.  The top level testbench file that instantiates the DUT must be named `tb.sv`.
+    The testbench module must be named `tb`, and the instance name of the DUT in
+    `tb.sv` must be `dut`. Using this standardized naming allows universal usage
+    of any supporting DV sources (coverage cfg files, and so on).
+2.  When instantiating any objects, the `name` argument to `type_id::create()`
+    must match the name of the variable being assigned to.
+    This ensures that any reporting messages refer to the right objects.
+3.  To avoid name collisions, it is recommended that global types and types defined
+    within a package have unique prefixes. This includes typedefs, enumerated
+    types/values, package parameters, package localparams, any functions/tasks, and
+    macros. This prefixing is not necessary for any types defined within a class,
+    module, or interface, as they will be scoped by the parent entity.
+4.  Note that all package variables (parameters, etc.) must have a specified type.
+5.  When instantiating a user-defined class member variable that is a class object,
+    it is recommended (but not necessary) to follow this prefixing scheme:
+    *   Prefix the instance name of the class object with `m_`, as such:
+
+        ```systemverilog
+        my_animal m_my_animal;
+        ```
+
+        This scheme does not apply to any class member variables that are
+        not themselves class objects.
+
+    *   If multiple instantiations of the class are desired without an array
+        to store them, use `<class_name> m_<instance_name>` to declare a variable,
+        and use an informative instance name.
+
+        ```systemverilog
+        my_animal m_my_cat;
+        my_animal m_my_dog;
+        my_animal m_my_fish;
+        ```
+
+    *   If an array of class handles or `logic`  is desired, choose a name such
+        that it follows one of the following schemes:
+
+        ```systemverilog
+        int NUM_ANIMALS = 10;
+        // This plural convention is recommended, but both are equally acceptable
+        my_animal m_my_animals[NUM_ANIMALS];
+        my_animal m_my_animal_array[NUM_ANIMALS];
+        ```
+
+
+## Directory Structure and Packages
+
+Define one class per file, and the filename should match the name of the class.
+The class files must be `` `include``-ed into the package file.
+Directory structure of the UVM agent and UVM environment testbench components
+along with their related files are shown below, with some additional guidance on
+declaring functions from a package context.
+
+
+
+
+### `uvm_agent` directory structure
+
+* All DV code for a particular agent will be placed under a single directory
+`<block>_agent/` separate from the DV code for the corresponding environment.
+* It is highly recommended that all agent directories be placed in a common area
+  to create a repository of UVM agents that allows for more vertical reuse.
+
+```systemverilog
+<block>_agent/
+  <block>_if.sv
+  <block>_item.sv
+  <block>_agent.sv
+  <block>_agent_pkg.sv
+  <block>_agent_cfg.sv
+  <block>_agent_cov.sv
+  <block>_driver.sv
+  <block>_host_driver.sv (Only if block requires a separate driver for host mode)
+  <block>_device_driver.sv (Only if block requires a separate driver for device mode)
+  <block>_monitor.sv
+  <block>_sequencer.sv
+  seq_lib/
+    <block>_base_seq.sv
+```
+
+
+### `uvm_env` directory structure
+* All DV code for the UVM environment, UVM tests, and top level block
+  testbench will be placed in a directory `<dut>/dv/` in three corresponding
+  subdirectories `<dut>/dv/env/, <dut>/dv/tests/, <dut>/dv/tb/`.
+* The `<dut>_env_pkg` should import the `<dut>_agent_pkg`.
+* All UVM virtual sequences for a given DUT will be placed under `env/seq_lib/`,
+  along with a `<dut>_vseq_list.sv` file that `` `include``-es all of the virtual
+  sequence files. This virtual sequence list file must be `` `include``-ed by
+  `env/<dut>_env_pkg.sv`.
+* If applicable, `env/<dut>_env_pkg.sv` must import the `<dut>_agent_pkg`.
+* `tests/<dut>_test_pkg.sv` must import the `<dut>_env_pkg` package.
+* All files `` `include``-ed into a single package must be in the same single directory.
+
+```systemverilog
+<dut>/dv/
+  env/
+    <dut>_env.sv
+    <dut>_env_pkg.sv
+    <dut>_env_cfg.sv
+    <dut>_env_cov.sv
+    <dut>_scoreboard.sv
+    <dut>_virtual_sequencer.sv
+    seq_lib/
+      // These allow basic functionality to be added to the environment
+      <dut>_base_vseq.sv
+      <dut>_sanity_vseq.sv
+  tb/
+    tb.sv
+  tests/
+    <dut>_base_test.sv
+    <dut>_test_pkg.sv
+```
+`<block>_agent_pkg.sv, <dut>_env_pkg.sv, <dut>_test_pkg.sv, and tb/tb.sv`
+  must include these `` `include``-es and imports:
+
+```systemverilog
+`include "dv_macros.svh"
+`include "uvm_macros.svh"
+import uvm_pkg::*;
+```
+
+An example environment level package file for an arbitrary block `newblock` should look similar to this:
+
+```systemverilog
+package newblock_env_pkg;
+  // dep packages
+  import uvm_pkg::*;
+  import newblock_agent_pkg::*;
+
+  // macro includes
+  `include "uvm_macros.svh"
+  `include "dv_macros.svh"
+
+  // define any parameters
+  parameter int NEW_BLOCK_PARAMETER = 10;
+
+  // define any local types/functions/tasks
+  typedef enum bit {
+    NewblockBooleanZero,
+    NewblockBooleanTrue
+  } newblock_boolean_e;
+
+  // Functions declared within packages must be automatic
+  function automatic bit newblock_test_function(newblock_boolean_e test_boolean);
+    //function definition
+  endfunction
+endpackage
+```
+
+
+## UVM Guidelines
+
+Always call `run_test()` without the test name argument. This allows command
+line test overrides to be performed using `+UVM_TESTNAME=<testname>` without
+having to recompile and allows testlists to be created for regression runs.
+
+
+### Class Definitions
+
+1.  After variable declarations, register any objects and components with the
+    factory using the `uvm_component_utils` and `uvm_object_utils` macros.
+    *   Declare any member variables to be registered with the UVM factory field
+    automation macros `uvm_field_int, uvm_field_enum, uvm_field_object,
+    uvm_field_array_int`, etc. within `uvm_object/component_utils_begin` and
+    `uvm_object/component_utils_end`.
+2.  For all field automation macros, use `UVM_DEFAULT` as the flag argument.
+    For example, if you want to enable everything except comparing and printing for
+    a `rx_delay` field, the macro would look like:
+
+    ```systemverilog
+    `uvm_field_int(rx_delay, UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
+    ```
+
+3.  After factory registration, a constructor must be defined. Use the
+    `uvm_object_new` macro for UVM objects and the `uvm_component_new` macro for UVM components.
+    Both of these macros are defined in `dv_macros.svh`.
+4.  Any constructors written manually should contain a call to
+    `super.new(name [,parent])` as the first line. Any other lines in the
+    constructor are optional.
+5.  Do not add any extra arguments to the constructor other than `name` and `parent`.
+6.  All objects should generally be instantiated in the `build_phase()` function
+    for components, or in the beginning of the `body()` task for sequences, unless
+    there is a good reason not to.
+    This is to enable proper usage of `uvm_factory` overrides.
+7.  When extending any user-defined component class, `super.<phase_name>_phase`
+    must be called for every phase method that is being overridden.
+8.  Transaction classes are usually logical representations of the data
+    communicated over one or more interfaces executing a particular protocol.
+    These classes:
+    *   Should extend `uvm_sequence_item`.
+    *   Should contain only information that is relevant to the data being
+        transmitted over an interface, and not to the means of transmit.
+    *   Should result in a legal transaction when randomized.
+
+
+### UVM `new()` functions
+
+The `new()` function of any class derived from `uvm_object` must have the
+default value of its `name` argument set to `""`, since the default name
+argument will be set by `type_id::create()`.
+The `new()` function of any class not derived from `uvm_object` must not
+have default values set for its arguments.
+Examples:
+
+:+1:
+```systemverilog
+// class derived from uvm_object with correct default value
+function new (string name = "");
+  super.new(name);
+endfunction : new
+
+// class derived from uvm_component with no default values
+function new(string name, uvm_component parent);
+  super.new(name, parent);
+endfunction : new
+```
+
+:-1:
+```systemverilog
+// Incorrect default value used for class derived from uvm_object
+function new(string name = "my_rand_seq");
+  super.new(name);
+endfunction : new
+
+// Incorrect because default values for class derived from uvm_component are provided
+function new(string name = "", uvm_component parent);
+  super.new(name, parent);
+endfunction : new
+```
+
+
+### `uvm_scoreboard` usage
+
+Use `uvm_tlm_analysis_fifo` - can be directly connected to analysis ports,
+and has an unbounded size so that writes will always succeed without blocking,
+in situations where transactions need to be stored for some time before being
+consumed by the scoreboard for processing.
+
+
+### `uvm_agent` usage
+
+1.  Interface agents should contain only a sequencer, driver, monitor, and config object.
+2.  Use one agent for every interface of the DUT, with an optional sequencer and
+    driver whose instantiation is determined by the value of `<agent_cfg>.is_active`.
+    This is placed in the agent config object to allow for better access throughout the testbench.
+3.  For any DUT interfaces that have multiple operation modes
+    (such as host and device modes), two drivers should be created, one for each mode,
+    and arbitration done in the `uvm_agent::build_phase` function to determine
+    which driver to instantiate.
+    Each driver should have different sequences that correspond to it.
+
+
+### `uvm_driver` usage
+
+1.  A driver must only communicate with one sequencer on one SystemVerilog
+    interface, and should not have any analysis ports.
+2.  A driver should not randomize any data received from transaction
+    items sent through analysis ports.
+3.  A driver must assign `X` to any buses it controls during "don't-care" clock
+    cycles in which no valid transaction is present.
+
+
+### Macro Usage
+
+The UVM library provides several macros as shortcuts for commonly used sequences
+of code. In addition to this, the provided `dv_macros.svh` file defines other
+utility macros for even more common code sequences.
+
+#### Guidelines for UVM library macros
+
+1.  All lowercase `` `uvm_... `` expressions denote library macros, and do not
+    require trailing semicolons.
+2.  `uvm_object_*` and `uvm_component_*` macros
+    *   Use these macros to register classes with the `uvm_factory`.
+3.  `uvm_field_*` macros
+    *   Use these macros for members of sequence items. While these macros
+        implement expensive versions of common functions required for sequence items,
+        the overhead is generally not worth implementing custom versions
+        for every sequence item. Additionally, these macros add a great deal of
+        readability and help avoid coding mistakes with custom implementations.
+        Only implement custom versions of these macros if performance becomes a
+        very noticeable issue. Do not use these macros with classes derived from `uvm_component`.
+4.  Do not use `uvm_do` macros. Instead, start sequence items on a sequencer
+    using the following high level steps:
+    *   `start_item(<sequence_item>, <priority>)`.
+    *   Randomize the sequence item.
+    *   `finish_item(<sequence_item>, <priority>)`.
+    *   `get_response(<response>)`.
+
+#### Guidelines for macros within `dv_macros.svh`
+
+1.  UVM print macros - `gfn`, `gtn`, `gn`, `gmv`
+    *   These macros provide a concise wrapper for `get_full_name()`,
+        `get_type_name()`, `get_name()`, and `uvm_reg.get_mirrored_value()`, respectively.
+        By using these it is possible to drastically reduce line length
+        and improve readability.
+2.  `downcast(<extended_class_handle>, <base_class_handle>)`
+    *   Use this macro to cast a handle for any base class object that
+        contains a handle to an extended class to the extended class type, useful
+        for situations that require polymorphism.
+3.  `uvm_object_new` and `uvm_component_new`
+    *   Use these macros to include the `new()` constructor for classes
+        derived from `uvm_object` and `uvm_component`, respectively.
+        If any additional logic must be included in the constructor,
+        these macros must not be used, and a constructor must be written manually.
+4.  `uvm_create_obj` and `uvm_create_comp`
+    *   Use these macros to call `create(<instance_name>)` and
+        `create(<instance_name>, this)`, respectively.
+5.  `DV_CHECK_*` comparison checking macros
+    *   These macros must be used to perform checks on the input arguments
+        and invoke one of the `uvm_info`, `uvm_error`, or `uvm_fatal`
+        reporting macros if the check fails.
+        These are much more concise than hand-writing the comparison checks and
+        the corresponding UVM report macro and provide efficient readability.
+    *   The severity of the invoked report macro can be specified with the
+        `<uvm_severity>` argument, the default reporting macro is `uvm_error`.
+    *   Append `_FATAL` to the end of these macro names to invoke variants of
+        these macros that will automatically throw fatal errors upon check failure
+        (e.g. `DV_CHECK_EQ_FATAL`, `DV_CHECK_LE_FATAL`, etc..).
+6.  Randomization macros
+    *   `dv_macros.svh` provides three sets of macros which must be used for
+        all randomization functionality:
+        *   `DV_CHECK_RANDOMIZE_FATAL` - Shorthand for `foo.randomize()`
+            with a `uvm_fatal` check.
+        *   `DV_CHECK_STD_RANDOMIZE_FATAL` - Shorthand for
+            `std::randomize(foo)` with a `uvm_fatal` check.
+        *   `DV_CHECK_MEMBER_RANDOMIZE_FATAL` - Shorthand for
+            `this.randomize(foo)` with a `uvm_fatal` check.
+    *   Variants of all three macros that allow constraints to be specified
+        also exist, these macros are invoked by inserting `WITH` into the macro name:
+        `DV_..._RANDOMIZE_WITH_FATAL`.
+7.  `DV_EOT_PRINT_..._CONTENTS` TLM fifo display macros
+    *   It is recommended to use these macros to display the status of the
+        various TLM fifos and queues in the scoreboard at the end of the test
+        to ensure that those data structures have been emptied. There should be no
+        pending transactions that have not yet been compared before the simulation
+        ends.
+        If used, these macros must be used during the `check_phase`.
+8.  `DV_SPINWAIT`
+    *   It is recommended to use this macro in situations that require an
+        isolation fork to disable some process after waiting for a specified
+        amount of time, as it provides proper process isolation such as the example
+        below:
+
+        ```systemverilog
+        // forked threads should be labeled if they serve a specific purpose,
+        // like the isolation thread below
+        fork : isolation_fork
+          begin
+            fork
+              begin
+                <specified code to execute>
+              end
+              begin
+                <watchdog timer logic with error reporting>
+              end
+            join_any
+            disable fork;
+          end
+        join
+        ```
+
+#### General macro use guidelines
+
+1.  Macro Naming:
+    *   If the macro wraps some fundamental UVM functionality and deals
+        directly with UVM functions, like `uvm_component_new`, and so on, the
+        macro must be named all in `lower_snake_case`.
+    *   If the macro expands into a block of code that does not deal directly
+        with UVM functionality, and is for convenience, like the
+        `DV_CHECK_*` macros, the macro must be named in `UPPER_SNAKE_CASE`.
+2.  Macro vs. Parameters:
+    *   If the intent is to define a constant that is a basic data type or
+        that is the result of an expression involving basic data types, a parameter
+        (or localparam) must be used, depending on where the constants are being defined.
+    *   If the intent is to abstract away a code snippet or to represent a
+        slice of an array, a macro must be used.
+        If the macro is defined in a local perspective, it must be undefined at
+        the end of the file, otherwise it will pollute the global namespace.
+
+
+### Factory
+
+To use the UVM factory, these guidelines should be followed:
+
+1.  All classes must be registered with the factory. Use the
+    `uvm_object_utils` and `uvm_component_utils` macros to register all
+    non-parameterized classes, and use `uvm_object_param_utils` and
+    `uvm_component_param_utils` to register all parameterized classes.
+2.  The `create()` API must be used to create all objects and components.
+    Exceptions for this rule are any TLM related objects and covergroup objects,
+    which should be created by directly calling `new()`, as they should not be
+    registered with the factory.
+3.  When using the factory to create objects with `<type_name>::type_id::create()`,
+    the `<type_name>` must be explicitly declared, parameters representing types
+    must not be used here.
+4.  When using the factory to create objects, it is recommended to use the
+    same name for both the name of the object and the variable used as a handle to
+    the object to prevent unnecessary confusion from error messages.
+
+    ```systemverilog
+    // create a single agent
+    m_myblock_agent = myblock_agent::type_id::create("m_myblock_agent", this);
+    ```
+
+5.  When using the factory to create an array of objects, decorate the name of
+    the object in the `create()` call with the index of the object in the array.
+
+    ```systemverilog
+    // create an array of objects
+    foreach (myobject_array[i]) begin
+      myobject_array[i] = myobject::type_id::create($sformatf("myobject_array[%0d]", i), this);
+    end
+    ```
+
+6.  For classes derived from `uvm_component`, the `parent` argument should be
+    the keyword `this`, so that the object hierarchy can be built correctly.
+
+    ```systemverilog
+    m_axi_driver = axi_driver::type_id::create("m_axi_driver", this);
+    ```
+
+7.  For classes derived from `uvm_object`, use `gfn` as part of the name
+    argument to append the full hierachy name, since `uvm_object` hierarchies are
+    not automatically built. Note that this only applies to class instances that
+    are not instantiated within a `uvm_component` in the test hierarchy,
+    as in this case the full hierarchy name will already be provided.
+
+    ```systemverilog
+    m_my_txn = my_txn::type_id::create({`gfn, ".m_my_txn"});
+    ```
+
+8.  All type or instance overrides must be in place before creating the class instance.
+    Factory usage examples:
+
+    ```systemverilog
+    // Override all drivers in testbench with my_driver
+    factory.set_type_override_by_type(current_driver::get_type(), my_driver::get_type());
+
+    // Override all drivers in testbench with my_driver - alternative syntax
+    current_driver::type_id::set_type_override(my_driver::get_type());
+
+    // Override specific instance of current_driver with my_driver
+    factory.set_inst_override_by_type("env1.m_agent1.driver", current_driver::get_type(), my_driver::get_type());
+
+    // Override specific instance of current driver - alternative syntax
+    current_driver::type_id::set_inst_override(my_driver::get_type(), "env.m_agent1.driver");
+    ```
+
+
+### Configuration Mechanism
+
+When using the UVM configuration database, these guidelines should be followed:
+
+1.  The `uvm_config_db` API must be used. Do not use `uvm_resource_db`. Do not use
+    `set/get_config_object()`, `set/get_config_string()`, or `set/get_config_int()`,
+    these have been deprecated.
+2.  In general, a designated configuration object extended from `uvm_object`
+    should be used for every `uvm_env` and `uvm_agent` component in the testbench.
+3.  It is recommended that the environment's configuration object contain the
+    agent's configuration object.
+4.  The `uvm_config_db` API should be used to pass configuration objects to
+    locations where they are needed. It should not be used to pass integers,
+    strings, or other basic data types, since it is much easier for namespace
+    collisions to occur when using lower level types.
+5.  You must always check for success when accessing the `uvm_config_db`.
+    If the lookup fails, issue a `uvm_fatal` to terminate the simulation.
+    This should be done as follows:
+
+    ```systemverilog
+    if (!uvm_config_db#(...)::get(...)) begin
+      `uvm_fatal(...)
+    end
+    ```
+6.  Wildcards must not be used in the field argument of `uvm_config_db::set(...)` calls.
+    It is allowed to use wildcards in the inst_name argument of these calls,
+    with proper precautions.
+
+
+### Sequences
+
+1.  It is recommended to keep sequences as generic as possible and to avoid
+    writing directed sequences unless absolutely necessary.
+2.  It is recommended to avoid explicit delays in terms of absolute time to
+    make code emulation friendly. Delays in terms of clock cycles are allowed.
+3.  The `body()` method should only execute the raw functional behavior of
+    the sequence.
+    Any associated housekeeping code should be placed elsewhere, such as in the
+    `pre_start()` and `post_start()` methods.
+    *   `pre_start()` and `post_start()` must be used instead of `pre_body()`
+        and `post_body()`, as these will always be called, even for subsequences
+        called from a parent sequence.
+4.  When creating a sequence object, always randomize the sequence before starting it.
+5.  Virtual sequences must be used to coordinate the behavior of multiple agents and multiple sequences.
+6.  Virtual sequences must be started on the null sequencer or on a valid virtual sequencer.
+
+
+### Sequence Items
+
+It is recommended, but not necessary, to create small "unit tests" for
+transactions objects to ensure the randomization constraints are working as
+intended.
+
+
+### RAL Usage
+
+When using the UVM RAL model, care must be taken when using `uvm_reg::set()` and
+`uvm_reg::update()`, as these are both non-atomic operations that can easily
+cause race conditions if there are any parallel threads.
+
+
+### Objections and Coordinating End of Test
+
+1.  At a high level, a phase should not end when either there is more stimulus
+    to send during that phase or the DUT has yet to respond to some stimulus that
+    has been sent.
+    To prevent a phase from ending in the former case, raise and lower objections
+    inside tests.
+    To prevent a phase from ending in the latter case, raise objections in the
+    scoreboard whenever expect queues are not empty and lower them when they are empty.
+2.  It is recommended to avoid using objections in drivers or monitors
+    for performance reasons they should be used inside these components
+    sparingly and only when needed.
+    (Note: this point is still under discussion, and is still TBD)
+3.  If a sequence needs to raise/drop objections, it should call
+    `raise_objection` in its `pre_start()` method and `drop_objection` in its
+    `post_start()` method. A good alternative to this is to call
+    `set_automatic_phase_objection(1)` in the sequence constructor.
+4.  Always perform the test `if (get_starting_phase() != null)` before calling
+    `raise_objection` or `drop_objection` within a sequence.
+5.  When starting a sequence that can raise and drop objections, if you want
+    the sequence to raise and drop objections, set the `starting_phase` member of
+    the sequence object the `<sequence_handle>.set_starting_phase(<uvm_phase>)`
+    before starting the sequence.
+6.  To avoid reimplementing objection functionality in multiple sequence classes,
+    create a parent sequence that implements this functionality for its derivatives.
+7.  When calling `raise_objection` or `drop_objection`, you must pass a string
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,6 +4,7 @@ This repository contains style guides curated by lowRISC for use in our
 code and documentation.
 
 - [SystemVerilog style guide](VerilogCodingStyle.md)
+- [SystemVerilog style guide for DV](VerilogForVerificationStyle.md)
```
cindychip: I thought this naming might be a little bit confusing. 
Is `SystemVerilog style guide` for RTL only or both DV and RTL?
What do you all think?

asb: I think the intent is that "SystemVerilog style guide" covers both as a baseline style guide, and "SystemVerilog style guide for DV" adds further guidance for DV. Maybe this could be presented in a more obvious way though...

udinator: Good point, perhaps this could be called something like `Design Verification Style Guide`, and omit SystemVerilog from that name entirely? I think the original guide can keep the name "SystemVerilog style guide".

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+assert(randomize(<variable>) with {<constraints>});
+// 3.
+assert(item.randomize() with {<constraints>});
+// 4.
+assert(function_that_has_side_effects());
+```
+
+### Wait and Fork
+
+Always put wait fork and disable fork constructs inside of an
```
vogelpi: Maybe you could put the ticks around "wait fork", "disable fork" and "isolation fork...join" to ease readability.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+these must be used to wrap relevant code, as shown below:
+
+```systemverilog
+`ifdef VCS
+  $stack();
+`endif
+```
+
+
+### Forbidden System Tasks and Functions
```
vogelpi: Maybe use "discouraged" instead of "forbidden"?

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+the sequence object the `<sequence_handle>.set_starting_phase(<uvm_phase>)`
+before starting the sequence.
+6. To avoid reimplementing objection functionality in multiple sequence classes,
+create a parent sequence that implements this functionality for its derivatives.
+7. When calling `raise_objection` or `drop_objection`, you must pass a string
+as a second argument to describe the objection to help with debugging.
+8. It is recommended, but not necessary, to use the built-in UVM timeout
+mechanism, `uvm_root::set_timeout(<time_limit>, 1)`.
+The base test in every testbench should specify a default timeout limit,
+which can be overriden by every derivative test.
```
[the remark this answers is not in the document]
udinator: Good catch, thanks.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+within a package have unique prefixes. This include typedefs, enumerated
+types/values, package parameters, package localparams, any functions/tasks, and
+macros.
+This prefixing is not necessary for any types defined within a class, as they
+will be scoped by the class.
+
+## Directory Structure and Packages
+
+*   Define one class per file, and the filename should match the name of the
+    class. The class files must be `included into the package file.
```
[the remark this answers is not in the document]
udinator: Fixed, thanks!

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+The UVM testbench generation script will automatically create skeletons of these
+components with this naming scheme, so almost all of this naming is automated
+out.
+Additional naming guidelines:
+
+1. When instantiating any objects, the `name` argument to `type_id::create()`
+must match the name of the variable being assigned to.
+This ensures that any reporting messages refer to the right objects.
+2. To avoid name collisions, it is recommended that global types and types defined
+within a package have unique prefixes. This include typedefs, enumerated
```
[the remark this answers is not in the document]
udinator: Fixed, thanks for pointing this out!

udinator:
> Thanks @udinator for driving this forward. I have a few suggestions regarding the actual content. Right now the styleguide contains many non-ASCII characters. These should be replaced by their ASCII counterparts. You can detect them with the following command
> 
> ```
> grep --color='auto' -P -n "[^\x00-\x7f]" VerilogForVerificationStyle.md
> ```

@vogelpi , thanks for the feedback and for pointing this out! I've replaced all of the non-ascii characters with their ascii equivalents.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+types/values, package parameters, package localparams, any functions/tasks, and
+macros. Note that all package variables (parameters, etc…) must have a specified type.
```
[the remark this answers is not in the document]
udinator: Done.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+An example environment level package file for an arbitrary block `newblock` should look similar to this:
+
+```systemverilog
+package newblock_env_pkg;
+  // dep packages
+  import uvm_pkg::*;
+  import newblock_agent_pkg::*;
+
+  // macro includes
+  `include “uvm_macros.svh”
```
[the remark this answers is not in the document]
udinator: Done, I replaced all non-ascii characters with the correct ascii representations.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+
+
+### Enums
+
+Enum names should be written in `lower_snake_case`, and the enumerated values
+should be written in `UpperCamelCase` as follows:
+
+```systemverilog
+// Enum type names should include the block name for clarity.
+// Enum value names should include the name of the enum.
```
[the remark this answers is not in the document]
udinator: That's a good point, I've moved these statements out of the code block to be standalone guidance before the code block.

VerilogForVerificationStyle.md
udinator: Yes, you're right, thanks for pointing this out!

VerilogForVerificationStyle.md
udinator: I feel that 'forbidden' would be slightly better for the purpose of this section, as the listed system functions must never be used in any lowRISC DV environment, as opposed to providing a recommendation to not use the functions. What do you think?

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+3. Use clocking blocks in a SystemVerilog interface to sample and drive
+synchronous DUT interfaces.
+4. It is recommended to implement a reusable interface scheme for block level
+testbenches that will be included in integration/system level testbenches.
+
+
+### Loop Operators
+
+It is highly recommended to use `foreach` loops as often as possible for
+iteration functionality, unless absolutely necessary.
```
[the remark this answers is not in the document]
udinator: Hmm, thanks for pointing this out. I removed the `absolutely necessary` part as it seemed somewhat redundant given that I provide some rationale in the very next sentence.

VerilogForVerificationStyle.md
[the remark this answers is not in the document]
vogelpi: Great, thanks.

VerilogForVerificationStyle.md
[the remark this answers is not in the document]
vogelpi: Perfect.

VerilogForVerificationStyle.md
vogelpi: In the design style guide we preferably use "discouraged". But I think there is no single true answer here. I guess "forbidden" is a very hard term. Question: Would you completely block the adoption/integration of code into a project if any of these functions are used? If yes, "forbidden" is the right term. If instead you would consider a step-wise process of integration first and at a later point rework the problematic parts, I would prefer "discouraged".

Anyway, I don't really mind. Please feel free to take the decision.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+// This plural convention is recommended, but both are equally acceptable
+my_animal my_animals[NUM_ANIMALS];
+my_animal my_animal_array[NUM_ANIMALS];
+```
+
+
+To avoid name collisions, it is recommended that global types and types defined
+within a package have unique prefixes. This include typedefs, enumerated
+types/values, package parameters, package localparams, any functions/tasks, and
+macros.
```
sriyerg: This is repeated word for word from bullet 2 above - perhaps make the bullet above more concise such that this elaborates more on it.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+This prefixing is not necessary for any types defined within a class, as they
```
sriyerg: class or module or interface....

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+  * [Macro Usage](#macro-usage)
+  * [Factory](#factory)
+  * [Configuration Mechanism](#configuration-mechanism)
+  * [Sequences](#sequences)
+  * [Sequence Items](#sequence-items)
+  * [Objections and Coordinating End of Test](#objections-and-coordinating-end-of-test)
+  * [Logging and Print Messages](#logging-and-print-messages)
+  * [DPI and C Connections](#dpi-and-c-connections)
+* [SystemVerilog Language Features](#systemverilog-language-features)
+  * [Randomization](#randomization)
```
sriyerg: Add a section on UVM RAL. Want to discourage the use of uvm_reg::set() followed by uvm_reg::update() since they are non-atomic operations which can cause race conditions if there are parallel threads.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+To avoid name collisions, it is recommended that global types and types defined
+within a package have unique prefixes. This include typedefs, enumerated
+types/values, package parameters, package localparams, any functions/tasks, and
+macros.
+This prefixing is not necessary for any types defined within a class, as they
+will be scoped by the class.
+
+## Directory Structure and Packages
+
+*   Define one class per file, and the filename should match the name of the
```
sriyerg: Is this mandatory or a recommendation? We have sources in our repo that violates this (`hw/dv/sv/tl_agent/tl_seq_lib.sv`, `hw/dv/sv/csr_utils/csr_seq_lib.sv`).

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+| Env level config object           | `rand <dut>_env_cfg cfg;`                    |
+| Env coverage collection object    | `<dut>_env_cov cov;`                         |
+| Agent                             | `<dut>_agent m_<dut>_agent;`                 |
+| Agent level config object         | `rand <dut>_agent_cfg m_<dut>_agent_cfg;`    |
+| Agent coverage collection object  | `<dut>_agent_cov cov;`                       |
+| Driver                            | `<dut>_driver driver;`                       |
+| Monitor                           | `<dut>_monitor monitor;`                     |
+| Scoreboard                        | `<dut>_scoreboard scoreboard;`               |
+| Virtual sequencer                 | `<dut>_virtual_sequencer virtual_sequencer;` |
+| Sequencer                         | `<dut>_sequencer <dut>_sequencer;`           |
```
sriyerg: Additional naming guidelines:
Testbench file that instantiates the DUT: `tb.sv`
Testbench module name: `tb`
Instance name of the DUT in `tb.sv`: `dut`

Standardizing these naming conventions makes it useful to create supporting DV sources that work universally across all IPs (example: `hw/dv/tools/vcs/cover.cfg` which tells VCS what coverage patterns to collect). Without the standard tb and dut names, each IP would have to create its own version which slightly increases effort.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+```
+
+* All DV code for a particular agent will be placed under a single directory
+`<block>_agent/` separate from the DV code for the corresponding environment.
+The `<block>_agent_pkg` will be imported by `<block>_env_pkg`
+* All DV code for the UVM environment, UVM tests, and top level block
+testbench will be placed in a directory `<block>/dv/` in three corresponding
+subdirectories `<block>/dv/env/, <block>/dv/tests/, <block>/dv/tb/`.
+* All UVM virtual sequences for a block will be placed under `env/seq_lib/`,
+along with a `<block>_vseq_list.sv` file that `includes all of the virtual
```
sriyerg: Wrap `includes in double back-ticks.

VerilogForVerificationStyle.md
tunghoang290780: Agree with @sriyerg. It'd better program the whole register via csr_wr instead of (set + update) combination

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+For example, if you want to enable everything except comparing and printing for
+a `rx_delay` field, the macro would look like:
+
+```systemverilog
+`uvm_field_int(rx_delay, UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
+```
+
+3.  After factory registration, a constructor must be defined. Use the
+`uvm_object_new` macro for UVM objects and the `uvm_component_new` macro for UVM components.
+Both of these macros are included in `dv_macros.svh`.
```
sriyerg: included -> defined

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+        `std::randomize(foo)` with a `uvm_fatal` check
+        *   `DV_CHECK_MEMBER_RANDOMIZE_FATAL` - Shorthand for
+        `this.randomize(foo)` with a `uvm_fatal` check
+    *   Variants of all three macros that allow constraints to be specified
+    also exist, these macros are invoked by inserting `WITH` into the macro name:
+    `DV_..._RANDOMIZE_WITH_FATAL`
+7. `DV_EOT_PRINT_..._CONTENTS` TLM fifo display macros
+    *   It is recommended to use these macros to display the status of the
+    various TLM fifos and queues in the scoreboard at the end of the test
+    to ensure correctness of the testbench checking mechanisms.
```
sriyerg: ....to ensure those data structures have been emptied. There should be no pending transactions that have not yet been compared before the simulation ends.

VerilogForVerificationStyle.md
sriyerg: 'Embargoed'?

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+  <block>_agent_cfg.sv
+  <block>_agent_cov.sv
+  <block>_driver.sv
+  <block>_host_driver.sv (Only if block requires a separate driver for host mode)
+  <block>_device_driver.sv (Only if block requires a separate driver for device mode)
+  <block>_monitor.sv
+  <block>_sequencer.sv
+  seq_lib/
+    <block>_base_seq.sv
+.hw/ip/<block>/dv/
```
tunghoang290780: Can `<block>_agent` and `hw/ip/<block>/dv` have different names. One IP (dut) may require multiple agents

sriyerg: Block agent and block env should be two separate topics / headings. They are not necessarily tied to each other (as Tung pointed out; plus, some blocks dont even have an agent). 

Also better to not make assumption on block env's directory path. It should be enough to indicate that it should reside alongside the DUT RTL code.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,826 @@
...
+
+## Directory Structure and Packages
+
+*   Define one class per file, and the filename should match the name of the
+    class. The class files must be `` `include``-ed into the package file.
+*   Directory structure of all relevant testbench files should follow the
+example below:
+
+```systemverilog
+<block>_agent/
```
sriyerg: Indicate that it is recommended to place the agent in a common area alongside other agents rather than with the DUT's UVM TB it was specifically written for - this creates a repository of UVC that users can find in one place.

VerilogForVerificationStyle.md
sriyerg: Maybe not call it 'block' agent - it can be any agent not necessarily tied to a particular block. Just call it 'interface' agent?

VerilogForVerificationStyle.md
udinator: Oops, you're right, I missed deleting this duplicate. I added the detail about prefixing types defined with a class to the bullet 2 above.

VerilogForVerificationStyle.md
udinator: Updated, thanks!

VerilogForVerificationStyle.md
udinator: Updated.

VerilogForVerificationStyle.md
udinator: Done.

VerilogForVerificationStyle.md
udinator: Thanks for the suggestion, that definitely sounds better. Updated.

VerilogForVerificationStyle.md
udinator: Hmm, shall we make an exception for sequence libraries and other similar libraries (e.g. in Ibex I use a library of `uvm_test` derivatives)? I think in the general case this should be a mandatory guideline, however.

VerilogForVerificationStyle.md
udinator: IMO, use of these functions should block adoption of code into a project, as these functions are either not in the SystemVerilog LRM or are not part of the SV randomization stability model (which could break reproducibility). I will add this rationale to the document.
What do you think @sriyerg?

VerilogForVerificationStyle.md
udinator: That is a good point, agree that the agent directory should be more tied to the interface that it is built around than the block in question, I'll update the naming to `<interface>_agent/` and split off the agent and env into two separate sections.

VerilogForVerificationStyle.md
udinator: Good idea, I've added this point.

VerilogForVerificationStyle.md
udinator: Updated, thanks for the suggestion!

VerilogForVerificationStyle.md
udinator: Agree, I think keeping the <block>_agent/ scheme and changing to <dut>_env would be the best option.

VerilogForVerificationStyle.md
udinator: This section has been added under the 'UVM Guidelines' heading, thanks for the suggestion!

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+`uvm_object_new` macro for UVM objects and the `uvm_component_new` macro for UVM components.
+Both of these macros are defined in `dv_macros.svh`.
+4.  Any constructors written manually should contain a call to
+`super.new(name [,parent])` as the first line. Any other lines in the
+constructor are optional.
+5.  Do not add any extra arguments to the constructor other than `name` and `parent`.
+6.  All objects should generally be instantiated in the `build_phase()` function
+for components, or in the beginning of the `body()` task for sequences, unless
+  there is a good reason not to.
+  This is to enable proper usage of uvm_factory overrides.
```
msfschaffner: you might want to tick this: uvm_factory - > `uvm_factory`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+and arbitration done in the `uvm_agent::build_phase` function to determine
+which driver to instantiate.
+Each driver should have different sequences that correspond to it.
+
+
+### `uvm_driver` usage
+
+1. A driver should only communicate with one sequencer on one SystemVerilog
+interface, and should not have any analysis ports.
+2. A driver should not randomize any data any data received from transaction
```
msfschaffner: ~~any data~~ any data

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+items sent through analysis ports.
+3. A driver must assign `X` to any buses it controls during "don't-care" clock
+cycles in which no valid transaction is present.
+
```
msfschaffner: Should we add a note here that contrasts this with what the RTL guide recommends in terms of usage of `X`?

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+// Override specific instance of current driver - alternative syntax
+current_driver::type_id::set_inst_override(my_driver::get_type(), "env.m_agent1.driver");
+```
+
+
+### Configuration Mechanism
+
+When using the UVM configuration database, these guidelines should be followed:
+
+1. The `uvm_config_db` API must be used. Do not `uvm_resource_db`. Do not use
```
msfschaffner: Do not **use** `uvm_resource_db`?

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+1. Print messages sent to the console should be kept to a minimum.
+Logs should be as concise as possible, by only writing key information when
+activity is present.
+2. Whenever possible, log any data directly to log files instead of the console.
+It is recommended that only critical information/errors be sent to both the
+console and the log files.
+3. Logging messages should be architected to be parsing "friendly"
+    *   All related information should be kept on a single line.
+    *   Only the relevant data should be logged, for example, there is no need
+    to log all fields of a transaction object when only a small subset are
```
msfschaffner: when only a small subset **is** valid

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+name for clarity.
+The enumerated values should be written in `UpperCamelCase` and should include
+the name of the enum for clarity.
+
+```systemverilog
+typedef enum bit {
+  UartInterruptFrameErr,
+  UartInterruptRxBufferFull,
+  ...
+} uart_interrupte_e;
```
msfschaffner: uart_interrupt~~e~~_e

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+[as described here](#macro-usage)
+
+
+### Void Casts
+
+Do not void cast any function/task calls that have useful return values,
+EXCEPT the following:
+* System functions
+* RAL `predict()` calls
+* FIFO/Queue methods
```
msfschaffner: maybe align to FIFO or fifo in the whole document for consistency

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+
+
+### Associative Arrays
+
+Do not use wildcard indexed [*] associative arrays.
```
msfschaffner: maybe use ticks around `[*]`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+## Introduction
+
+This guide defines the Comportable style for DV-specific SystemVerilog code. The
+goal is to use this common set of guidelines when implementing testbenches using
+the UVM methodology to allow all developed Comportable DV code to be uniform,
+reusable, and portable.
+This document assumes previous experience and working knowledge of the UVM, and
+as such all included code snippets will be fairly concise.
+A testbench generation script
+[uvmdvgen.py](https://github.com/lowRISC/opentitan/blob/master/util/uvmdvgen.py)
```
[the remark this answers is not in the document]
udinator: @sriyerg and I had a brief discussion regarding this, and this script will likely follow the dv base libraries into the separate lowRISC IP repository. For the time being, I refer to this script in the style guide just to provide another option for testbench generation, especially because much of the testbench style guidelines in the document are mirrored in the auto-generated UVM files produced by this script.

VerilogForVerificationStyle.md
udinator: Yep, good catch!

VerilogForVerificationStyle.md
udinator: Fixed, thanks for the pointer!

VerilogForVerificationStyle.md
udinator: I feel that may not be needed, since the RTL guide explictly states that the `X` guidelines should apply to RTL designs only, whereas here we only refer to code that lives within the UVM framework. What do you think?

VerilogForVerificationStyle.md
udinator: Yep, you're right, good catch

VerilogForVerificationStyle.md
udinator: Fixed.

VerilogForVerificationStyle.md
udinator: Fixed.

VerilogForVerificationStyle.md
udinator: Good point, I changed this usage to lowercase to match the rest of the document.

VerilogForVerificationStyle.md
udinator: Updated.

udinator:
> Thanks @udinator! I left a few minor comments here and there, but overall LGTM

Thanks for the comments and feedback @msf! Hopefully I should have addressed all of the sections that you pointed out.

VerilogForVerificationStyle.md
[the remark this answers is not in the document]
msfschaffner: SGTM, thanks

VerilogForVerificationStyle.md
msfschaffner: That is true, it explicitly states that the guidance applies to RTL. Ok if we omit it here.

msfschaffner:
Thanks for addressing my points, @udinator. LGTM

VerilogForVerificationStyle.md
GregAC: I think forbidden can be justified in some cases, in particular the `$random` and `$dist_*` mentioned below which can break simulation reproducibility. Breaking reproducibility can be very painful and hard to diagnose so this seems a good reason to forbid those functions.

The others don't seem to cause such a serious problem if they occur but as they're not part of the LRM could break the build if you have a simulator that doesn't support them, again a decent reason to forbid them.

sriyerg: Agree - we may have much control over 3rd party code, but if they are using these, then we file bugs and get them to fix it. We do want to forbid the active contributors of OpenTitan from using these. $psprintf is problematic too since its not a part of the LRM and some simulators may not support it.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+  // define any parameters
+  parameter int MY_BLOCK_PARAMETER = 10;
+
+  // define any local types/functions/tasks
+  typedef enum bit {
+    MyblockBooleanZero,
+    MyblockBooleanTrue
+  } myblock_boolean_e;
+
+  // Functions declared within packages must be automatic
```
sriyerg: We need to expand this to include all other static entities (modules and interfaces). Also, reword this to indicate that the scope should be explicitly indicated, whether automatic or static (when in doubt, use automatic). We may have usecases for static functions as well.

VerilogForVerificationStyle.md
weicaiyang: Agree. IMO, it's good to forbid these items.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,850 @@
...
+components with this naming scheme, so almost all of this naming is automated
+out.
+Additional naming guidelines:
+
+1. The top level testbench file that instantiates the DUT must be named `tb.sv`.
+   The testbench module must be named `tb`, and the instance name of the DUT in
+   `tb.sv` must be `dut`. Using this standardized naming allows universal usage
+   of any supporting DV sources (coverage cfg files, and so on).
+2. When instantiating any objects, the `name` argument to `type_id::create()`
+must match the name of the variable being assigned to.
```
sriyerg: nit: alignment of following lines

VerilogForVerificationStyle.md
udinator: Good point, I spun this off into a separate small section under the Packages heading.

VerilogForVerificationStyle.md
udinator: Thanks for the pointer, I've fixed this alignment of number/bullet lists throughout the document.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    *   It is recommended to use these macros to display the status of the
+    various TLM fifos and queues in the scoreboard at the end of the test
+    to ensure that those data structures have been emptied. There should be no
+    pending transactions that have not yet been compared before the simulation
+    ends.
+    If used, these macros must be used during the `check_phase`.
+8.  `DV_SPINWAIT`
+    *   It is recommended to use this macro in situations that require an
+    isolation fork to disable some process after waiting for a specified
+    amount of time, as it provides a built in timer watchdog and proper process isolation:
```
sjgitty: recommend "proper process isolation such as the example below."

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    join_any
+    disable fork;
+  end
+join
+```
+
+General macro use guidelines:
+
+1.  Macro naming
+    *   If the macro wraps some fundamental UVM functionality and deals
```
sjgitty: trailing `.`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+      begin
+        <watchdog timer logic with error reporting>
+      end
+    join_any
+    disable fork;
+  end
+join
+```
+
+General macro use guidelines:
```
sjgitty: `#### General macro use guidelines`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    non-parameterized classes, and use `uvm_object_param_utils` and
+    `uvm_component_param_utils` to register all parameterized classes.
+2.  The `create()` API must be used to create all objects and components.
+    Exceptions for this rule are any TLM related objects and covergroup objects,
+    which should be created by directly calling `new()`, as they should not be
+    registered with the factory.
+3.  When using the factory to create objects with `<type_name>::type_id::create()`,
+    the `<type_name>` must be explicitly declared, parameters representing types
+    must not be used here.
+3.  When using the factory to create objects, it is recommended to use the
```
sjgitty: Evidently `.md` will heed your numbering when you pick up a new numbered list
but autoincrements for you if it is a continuation of a numbered list. So this `3`
renders as `4` (correctly), but the next `4` renders as `4` when it should be `5`.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    same name for both the name of the object and the variable used as a handle to
+    the object to prevent unnecessary confusion from error messages.
+
+```systemverilog
+// create a single agent
+m_myblock_agent = myblock_agent::type_id::create("m_myblock_agent", this);
+```
+
+4.  When using the factory to create an array of objects, decorate the name of
```
sjgitty: should be `5.`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    the object in the `create()` call with the index of the object in the array.
+
+```systemverilog
+// create an array of objects
+foreach (myobject_array[i]) begin
+  myobject_array[i] = myobject::type_id::create($sformatf("myobject_array[%0d]", i), this);
+end
+```
+
+5.  For classes derived from `uvm_component`, the `parent` argument should be
```
sjgitty: should be `6.`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    the keyword `this`, so that the object hierarchy can be built correctly.
+
+```systemverilog
+m_axi_driver = axi_driver::type_id::create("m_axi_driver", this);
+```
+
+6.  For classes derived from `uvm_object`, use `gfn` as part of the name
```
sjgitty: `7.`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    argument to append the full hierachy name, since `uvm_object` hierarchies are
+    not automatically built. Note that this only applies to class instances that
+    are not instantiated within a `uvm_component` in the test hierarchy,
+    as in this case the full hierarchy name will already be provided.
+
+```systemverilog
+m_my_txn = my_txn::type_id::create({`gfn, ".m_my_txn"});
+```
+
+7.  All type or instance overrides must be in place before creating the class instance.
```
sjgitty: `8.`

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+2.  It is recommended to avoid explicit delays in terms of absolute time to
+    make code emulation friendly. Delays in terms of clock cycles are allowed.
+3.  The `body()` method should only execute the raw functional behavior of
+    the sequence.
+    Any associated housekeeping code should be placed elsewhere, such as in the
+    `pre_start()` and `post_start()` methods.
+    *   `pre_start()` and `post_start()` must be used instead of `pre_body()`
+    and `post_body()`, as these will always be called, even for subsequences
+    called from a parent sequence.
+5.  When creating a sequence object, always randomize the sequence before starting it.
```
sjgitty: It does the right thing here, but should probably fix it to `4` in case
you ever break up the list with ```code blocks``` in the future

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    *   Declare any member variables to be registered with the UVM factory field
+    automation macros `uvm_field_int, uvm_field_enum, uvm_field_object,
+    uvm_field_array_int`, etc. within `uvm_object/component_utils_begin` and
+    `uvm_object/component_utils_end`.
+2.  For all field automation macros, use `UVM_DEFAULT` as the flag argument.
+    For example, if you want to enable everything except comparing and printing for
+    a `rx_delay` field, the macro would look like:
+
+```systemverilog
+`uvm_field_int(rx_delay, UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
```
sjgitty: probably want a reminder early here or somewhere that `uvm_<name> is a macro that doesn't require a trailing semicolon

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+
+
+
+## Introduction
+
+This guide defines the Comportable style for DV-specific SystemVerilog code. The
+goal is to use this common set of guidelines when implementing testbenches using
+the UVM methodology to allow all developed Comportable DV code to be uniform,
+reusable, and portable.
+This document assumes previous experience and working knowledge of the UVM, and
```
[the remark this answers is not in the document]
udinator: Sure, I don't have any problem with changing it to `of UVM`, it's been updated.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+as such all included code snippets will be fairly concise.
+A testbench generation script
```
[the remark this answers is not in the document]
udinator: I agree, I'm also starting to feel like these references would be better off just removed from the document altogether since it somewhat gives the impression that the guide is based off of the format of the testbench generation script, instead of the other way around.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+| Agent                             | `<dut>_agent m_<dut>_agent;`                 |
+| Agent level config object         | `rand <dut>_agent_cfg m_<dut>_agent_cfg;`    |
+| Agent coverage collection object  | `<dut>_agent_cov cov;`                       |
+| Driver                            | `<dut>_driver driver;`                       |
+| Monitor                           | `<dut>_monitor monitor;`                     |
+| Scoreboard                        | `<dut>_scoreboard scoreboard;`               |
+| Virtual sequencer                 | `<dut>_virtual_sequencer virtual_sequencer;` |
+| Sequencer                         | `<dut>_sequencer <dut>_sequencer;`           |
+
+The UVM testbench generation script will automatically create skeletons of these
```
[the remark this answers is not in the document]
udinator: As per my response in your previous comment, I've deleted this section, just to simplify things as much as possible.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    module, or interface, as they will be scoped by the parent entity.
+4.  Note that all package variables (parameters, etc.) must have a specified type.
+5.  When instantiating a user-defined class member variable that is a class object:
+    *   Prefix the instance name of the class object with `m_`, as such:
+
+```systemverilog
+my_animal m_my_animal;
+```
+
+This scheme does not apply to any class member variables that are
```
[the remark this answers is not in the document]
udinator: Hmm, there should be a way to fix this issue, I'll look into it. Thanks for pointing that out!

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+not themselves class objects.
+
+* If multiple instantiations of the class are desired without an array
+to store them, use `<class_name> m_<instance_name> to declare a variable,
```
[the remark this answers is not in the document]
udinator: Fixed

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+my_animal m_my_fish;
+```
+
+* If an array of class handles is desired, choose a name such that it follows
+one of the following schemes:
+
+```systemverilog
+int NUM_ANIMALS = 10;
+// This plural convention is recommended, but both are equally acceptable
+my_animal my_animals[NUM_ANIMALS];
```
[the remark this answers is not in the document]
udinator: Added as a mandatory guideline.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+modules and interfaces, must be explicitly scoped with either the `static` or
+`automatic` keyword.
+
+
+### `uvm_agent` directory structure
+
+* All DV code for a particular agent will be placed under a single directory
+`<block>_agent/` separate from the DV code for the corresponding environment.
+* It is highly recommended that all agent directories be placed in a common area
+  to create a common agent UVC repository that allows for more vertical reuse.
```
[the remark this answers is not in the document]
udinator: Generally speaking it should be, but I can expand it here for clarity.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+  seq_lib/
+    <block>_base_seq.sv
+```
+
+
+### `uvm_env` directory structure
+* All DV code for the UVM environment, UVM tests, and top level block
+  testbench will be placed in a directory `<dut>/dv/` in three corresponding
+  subdirectories `<dut>/dv/env/, <dut>/dv/tests/, <dut>/dv/tb/`.
+* The `<dut>_env_pkg` should import the `<dut>_agent_pkg`
```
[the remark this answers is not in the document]
udinator: Fixed

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+2. `uvm_field_*` macros
+    *   Use these macros for members of sequence items. While these macros
+    implement expensive versions of common functions required for sequence items,
+    the overhead is generally not worth implementing custom versions
+    for every sequence item. Additionally, these macros add a great deal of
+    readability and help avoid coding mistakes with custom implementations.
+    Only implement custom versions of these macros if performance becomes a
+    very noticeable issue. Do not use these macros with classes derived from `uvm_component`.
+3. `uvm_do` macros for executing sequences and sequence items
+    *   Do not use these.
```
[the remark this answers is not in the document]
udinator: Good point, updated.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    cycles in which no valid transaction is present.
+
+
+### Macro Usage
+
+The UVM library provides several macros as shortcuts for commonly used sequences
+of code. In addition to this, the provided `dv_macros.svh` file defines other
+utility macros for even more common code sequences.
+
+We shall list guidelines for the UVM library macros first:
```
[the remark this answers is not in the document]
udinator: Makes sense, will update table of contents accordingly.

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    very noticeable issue. Do not use these macros with classes derived from `uvm_component`.
+3. `uvm_do` macros for executing sequences and sequence items
+    *   Do not use these.
+    *   Execute sequence items using the following high level steps:
+        *   `start_item(<sequence_item>, <priority>)`
+        *   randomize the sequence item
+        *   `finish_item(<sequence_item>, <priority>)`
+        *   `get_response(<response>)`
+
+We shall now list some guidelines for the macros provided in `dv_macros.svh`:
```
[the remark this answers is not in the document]
udinator: Updated

VerilogForVerificationStyle.md
```diff
@@ -0,0 +1,863 @@
...
+    *   Use these macros to call `create(<instance_name>)` and
+    `create(<instance_name>, this)`, respectively.
+5.  `DV_CHECK_*` comparison checking macros
+    *   These macros must be used to perform checks on the input arguments
+    and invoke one of the `uvm_info`, `uvm_error`, or `uvm_fatal`
+    reporting macros if the check fails.
+    These are much more concise than hand-writing the comparison checks and
+    the corresponding UVM report macro and provide efficient readability.
+    *   The severity of the invoked report macro can be specified with the
+    <uvm_severity> argument, the default reporting macro is `uvm_error`.
```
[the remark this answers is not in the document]
udinator: Updated

VerilogForVerificationStyle.md
udinator: Thanks for the pointer, fixed.

VerilogForVerificationStyle.md
udinator: Updated

VerilogForVerificationStyle.md
udinator: Ah yeah, seems I accidentally listed two items as `3.`, I've updated this list to fix the issue.

VerilogForVerificationStyle.md
udinator: Good catch, fixed

VerilogForVerificationStyle.md
udinator: Good idea, I included a brief note about this in the UVM library macros section.

DVCodingStyle.md
```diff
@@ -0,0 +1,863 @@
...
+along with their related files are shown below, with some additional guidance on
+declaring functions from a package context.
+
+
+### Function Declarations
+
+Functions declared within packages, as well as any other static entities such as
+modules and interfaces, must be explicitly scoped with either the `static` or
+`automatic` keyword.
+
```
sriyerg: Making this a sub-heading of "Directory Structure" seems a little strange.

DVCodingStyle.md
```diff
@@ -0,0 +1,863 @@
+# lowRISC SystemVerilog Coding Style Guide for Design Verification
+
+This document is a supplemental Style Guide to the [lowRISC Verilog style
+guide](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md),
+with emphasis on writing code for Design Verification (DV) in SystemVerilog,
+following the UVM methodology.
```
sriyerg: Link the UVM methodology to: 
https://www.accellera.org/images//downloads/standards/uvm/uvm_users_guide_1.2.pdf

sriyerg: [Or indicate this in the introduction]

DVCodingStyle.md
```diff
@@ -0,0 +1,863 @@
...
+        my_animal m_my_animals[NUM_ANIMALS];
+        my_animal m_my_animal_array[NUM_ANIMALS];
+        ```
+
+
+## Directory Structure and Packages
+
+Define one class per file, and the filename should match the name of the class.
+The class files must be `` `include``-ed into the package file.
+Directory structure of the `uvm_agent` and `uvm_env` testbench components
```
sriyerg: nit: UVM agent and UVM environment instead of `uvm_agent` and `uvm_env`

VerilogForVerificationStyle.md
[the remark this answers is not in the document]
sriyerg: Agree - our goal should be to make this generic enough that anyone who finds it on the internet is able to apply it for their project.

DVCodingStyle.md
udinator: I think a link would look better, added.

DVCodingStyle.md
udinator: Fixed

DVCodingStyle.md
udinator: True, it does make sense to move this to under the SV language features heading instead.
Thanks for pointing that out!

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+  // macro includes
+  `include "uvm_macros.svh"
+  `include "dv_macros.svh"
+
+  // define any parameters
+  parameter int MY_BLOCK_PARAMETER = 10;
+
+  // define any local types/functions/tasks
+  typedef enum bit {
+    MyblockBooleanZero,
```
sjgitty: What's the relationship between newblock and Myblock, or should that be NewblockBoolean...

sjgitty:
would love to see a review from @rasmus-madsen but otherwise I think this is almost ready to go. Thanks @udinator

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+
+### Objections and Coordinating End of Test
+
+1.  At a high level, a phase should not end when either there is more stimulus
+    to send during that phase or the DUT has yet to respond to some stimulus that
+    has been sent.
+    To prevent a phase from ending in the former case, raise and lower objections
+    inside tests.
+    To prevent a phase from ending in the latter case, raise objections in the
+    scoreboard whenever expect queues are not empty and lower them when they are empty.
```
rasmus-madsen: I completely disagree with the comments in line 602 og 603.
I would say to avoid using objections in scoreboards and use them in monitors instead.

Raising objections in scoreboard can lead to deadlocks.  Ideally objections should only be in the test. 
but if needed a monitor can raise and objection in the Phase_ready_to_end phase if data is seen.

Also raising objections outside of a test should only be done in the phase_ready_to_end to improve performance.

maybe the topic of objections warrants separate thread?

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+### Function Declarations
+
+Functions declared within packages, as well as any other static entities such as
+modules and interfaces, must be explicitly scoped with either the `static` or
+`automatic` keyword.
+
+### Randomization
+
+Good constraint coding style is important for faster simulations and better
+simulation performance.
```
rasmus-madsen: I would suggest we recommend using a constraint name that matches what is being constrained.
 for the example below I would suggest a naming scheme with either

a_con
con_a
or 
c_a

rasmus-madsen:
@senelson7 maybe you want to look at this too

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+  function automatic bit myblock_test_function(myblock_boolean_e test_boolean);
+    //function definition
+  endfunction
+endpackage
+```
+
+
+## UVM Guidelines
+
+Always call `run_test()` without the test name argument. This allows command
```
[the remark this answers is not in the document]
sjgitty: > nit:
> I think the mark down style guide dictates only one sentence per line
> and no line breaks in a sentence

Unfortunately this document falls in the category of "unrendered / README files"
in the [markdown style guide](https://docs.opentitan.org/doc/rm/markdown_usage_style/#line-length)
and thus has to stay at 80 characters wrapped rather than one sentence per line.

DVCodingStyle.md
udinator: Good catch, let me change this to NewblockBoolean.

sjgitty:
> @senelson7 maybe you want to look at this too

Thanks Rasmus for the input. There will be some style and philosophical
disagreements for sure, but it is great to get your opinions on it. @senelson7 
please weigh in too.

@sriyerg @udinator see if you have good counter arguments for some of
the issues @rasmus-madsen has raised.

DVCodingStyle.md
udinator: That is a good point you bring up, the style that we have generally used so far is `<constraint_name>_c`, let us standardize that format if that sounds good to you, and I'll add it to the style guide?
@sriyerg @cindychip @weicaiyang what do you think?

cindychip: Sure I agree with adding `<constraint_name>_c`

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+        ```
+
+    *   If an array of class handles is desired, choose a name such that it follows
+        one of the following schemes:
+
+        ```systemverilog
+        int NUM_ANIMALS = 10;
+        // This plural convention is recommended, but both are equally acceptable
+        my_animal m_my_animals[NUM_ANIMALS];
+        my_animal m_my_animal_array[NUM_ANIMALS];
```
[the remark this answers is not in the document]
sriyerg: This is also just a recommendation.

DVCodingStyle.md
sriyerg: I see some valid usecases in allowing raising/dropping objections in the scoreboard; example: in xbar type DUT, raise obj when transaction enters the DUT and drop only when it exits. This objection (that helps ensure that the test does not exit prematurely) cannot be captured in the monitor and has to be captured in a higher level component. It cannot be captured in the test either since it is an entity that is only generating and bombarding C-R stimulus into the DUT. It is no knowledge of when the transactions enter and exit the DUT. 

We keep the performance impact low by raising objection exactly once (dont raise if it is already raised), which is what the line implies. You can find that logic in our base scoreboard (`hw/dv/sv/dv_lib/dv_base_scoreboard.sv`). If by deadlocks you mean hangs, then there is an obvious problem either in the design or dv, that needs fixing. We avoid hanging forever through reasonable test timeouts.

I have not used `phase_ready_to_end` and would like to learn more about it. Let's discuss this in a separate thread.

DVCodingStyle.md
weicaiyang: agree with <constraint_name>_c, too

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+    must match the name of the variable being assigned to.
+    This ensures that any reporting messages refer to the right objects.
+3.  To avoid name collisions, it is recommended that global types and types defined
+    within a package have unique prefixes. This includes typedefs, enumerated
+    types/values, package parameters, package localparams, any functions/tasks, and
+    macros. This prefixing is not necessary for any types defined within a class,
+    module, or interface, as they will be scoped by the parent entity.
+4.  Note that all package variables (parameters, etc.) must have a specified type.
+5.  When instantiating a user-defined class member variable that is a class object:
+    *   Prefix the instance name of the class object with `m_`, as such:
```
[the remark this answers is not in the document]
rasmus-madsen: @sriyerg thank you for the examples
I see your point. but that raises a different question.
Is the style guide to help the coder or the maintainer (which could be the same person)
I think it is to get the coder to write code in such a way it is easy to understand and maintain.

The coder should know exactly what they are doing - and can use any prefix as they will know what it means.
The maintainer might come from a different world and have different assumptions what m_ and other post/pre fixes indicate.
as mentioned I've seen m_ / s_ used as indication of master and slave driven parts of a bus.

while I do have an opinion, I am mostly playing the devils advocate here to make sure we weight pro's and cons before making decisions

DVCodingStyle.md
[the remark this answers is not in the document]
rasmus-madsen: that is fine.. I am also just pointing out that 
a bus is technically an array of logic - which should be included if we really want this.
and that don't see any added value as the instantiation will show that this is an array.

DVCodingStyle.md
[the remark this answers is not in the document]
rasmus-madsen: @sjgitty not sure that is unfortunate.. ;)
@udinator  sorry for missing that fact.

DVCodingStyle.md
rasmus-madsen: I am ok with _c also

DVCodingStyle.md
rasmus-madsen: I have created #30 do discuss further

DVCodingStyle.md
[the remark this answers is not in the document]
sriyerg: Sure, we can update the style guide to add it as well.

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+    for performance reasons.
+3.  If a sequence needs to raise/drop objections, it should call
+    `raise_objection` in its `pre_start()` method and `drop_objection` in its
+    `post_start()` method. A good alternative to this is to call
+    `set_automatic_phase_objection(1)` in the sequence constructor.
+4.  Always perform the test `if (starting_phase != null)` before calling
+    `raise_objection` or `drop_objection` within a sequence. Prior to uvm-1.2,
+    `starting_phase` was a member of the class `uvm_sequence_base`.
+    From uvm-1.2 onward, the `starting_phase` variable is deprecated and instead
+    must be accessed using the `get_starting_phase` method.
```
rswarbrick: Do we have a version of UVM that we'll always target? If >= 1.2, this can be simplified.

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+    before starting the sequence.
+6.  To avoid reimplementing objection functionality in multiple sequence classes,
+    create a parent sequence that implements this functionality for its derivatives.
+7.  When calling `raise_objection` or `drop_objection`, you must pass a string
+    as a second argument to describe the objection to help with debugging.
+8.  It is recommended, but not necessary, to use the built-in UVM timeout
+    mechanism, `uvm_root::set_timeout(<time_limit>, 1)`.
+    The base test in every testbench should specify a default timeout limit,
+    which can be overridden by every derivative test.
+    A standard plusarg is recommended to be used for this override mechanism.
```
rswarbrick: Can we name the plusarg here?

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+    From uvm-1.2 onward, the `starting_phase` variable is deprecated and instead
+    must be accessed using the `get_starting_phase` method.
+5.  When starting a sequence that can raise and drop objections, if you want
+    the sequence to raise and drop objections, set the `starting_phase` member of
+    the sequence object the `<sequence_handle>.set_starting_phase(<uvm_phase>)`
+    before starting the sequence.
+6.  To avoid reimplementing objection functionality in multiple sequence classes,
+    create a parent sequence that implements this functionality for its derivatives.
+7.  When calling `raise_objection` or `drop_objection`, you must pass a string
+    as a second argument to describe the objection to help with debugging.
```
rswarbrick: This is sensible, but can we give examples? In particular, it's a bit less clear what sensible strings for `drop_objection` should be.

sjgitty:
Are we close to landing this? what are the remaining steps? We can always
amend as needed once landed.

udinator:
> Are we close to landing this? what are the remaining steps? We can always
> amend as needed once landed.

@sjgitty - There are a few minor updates that I need to make, corresponding to a few open comments in this PR, and a few bigger issues (end of test mechanism, and reactive agent usage).
@sriyerg , how would you feel if I added the sections about EoT using `phase_ready_to_end` and reactive agents in a separate PR, just to get the initial guide landed?

DVCodingStyle.md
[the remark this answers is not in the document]
udinator: @sriyerg @rasmus-madsen , based on some discussion we had about this point, we think making this prefixing optional would be a good way to go

DVCodingStyle.md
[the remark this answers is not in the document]
udinator: Updated to `If an array of class handles or logic is desired`.

DVCodingStyle.md
```diff
@@ -0,0 +1,864 @@
...
+    and arbitration done in the `uvm_agent::build_phase` function to determine
+    which driver to instantiate.
+    Each driver should have different sequences that correspond to it.
+
+
+### `uvm_driver` usage
+
+1.  A driver should only communicate with one sequencer on one SystemVerilog
+    interface, and should not have any analysis ports.
+2.  A driver should not randomize any data received from transaction
```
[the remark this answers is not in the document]
udinator: Updated

DVCodingStyle.md
udinator: @sriyerg what are your thoughts about this phrasing?

DVCodingStyle.md
udinator: yep good point, updated

DVCodingStyle.md
udinator: Sure, this has been added.

DVCodingStyle.md
sriyerg: At the moment we are using UVM 1.2 with both simulators, so wouldn't this be flagged as a compile error (and hence force the writer to update the code anyway)? I don't think we will ever revert back to an older version.

sriyerg:
> > Are we close to landing this? what are the remaining steps? We can always
> > amend as needed once landed.
> 
> @sjgitty - There are a few minor updates that I need to make, corresponding to a few open comments in this PR, and a few bigger issues (end of test mechanism, and reactive agent usage).
> @sriyerg , how would you feel if I added the sections about EoT using `phase_ready_to_end` and reactive agents in a separate PR, just to get the initial guide landed?

Separate PR SGTM.

rasmus-madsen:
@sriyerg is it on purpose that I have merge permissions in this repository?

rasmus-madsen:
> > Are we close to landing this? what are the remaining steps? We can always
> > amend as needed once landed.
> 
> @sjgitty - There are a few minor updates that I need to make, corresponding to a few open comments in this PR, and a few bigger issues (end of test mechanism, and reactive agent usage).
> @sriyerg , how would you feel if I added the sections about EoT using `phase_ready_to_end` and reactive agents in a separate PR, just to get the initial guide landed?

Can we at least make a comment in the styleguide that this section is still TDB/WIP

sriyerg:
> @sriyerg is it on purpose that I have merge permissions in this repository?

I think this is just style guides repo - there is nothing critical here to prevent anyone from contributing to these docs. The project/repo admins (LR) can answer this better.

DVCodingStyle.md
udinator: Updated to solely reflect usage of the `get_starting_phase()` method.

udinator:
> Can we at least make a comment in the styleguide that this section is still TDB/WIP

Sure @rasmus-madsen - I've added a small note that the EoT points are still under discussion.

udinator:
@sjgitty , @sriyerg , if you are okay with the current state of the style guide, I can merge it into the repo, and make amendments in the future as necessary.

sjgitty:
LGTM, I'd say let's ship it and amend where necessary

Issue #16, opened 2020-01-14
Dealing with 'unavoidable' Xs to case statement inputs


GregAC:
This Ibex issue: https://github.com/lowRISC/ibex/issues/548 has raised the question of how we should deal with Xs going into case statements where we cannot avoid that X. From the style guide a typical case statement should look something like

```verilog
`ASSERT_KNOWN(CaseInputKnown, case_input, clk_i, !rst_ni)
always_comb begin
  case_output = 4'b0000;
  unique case (case_input)
    2'b00: case_output = 4'b0001;
    2'b01: case_output = 4'b0010;
    2'b10: case_output = 4'b0100;
    2'b11: case_output = 4'b1000;
    default: ;
  endcase
end
```

How should we deal with the case where `case_input` could be X? e.g. it could come direct from a top-level input that the DV environment may drive to X or from the output of a memory.

In such cases there's generally some kind of valid signal (`case_valid`) that lets us know that `case_input` can be ignored and hence `case_output` should be ignored.

You could say `case_input` must be qualified by `case_valid` to remove the X from the case input

```verilog
logic [1:0] case_input_qualified;

assign case_input_qualified = case_input & {2{case_valid}};
`ASSERT_KNOWN(CaseInputKnown, case_input_qualified, clk_i, !rst_ni)
always_comb begin
  case_output = 4'b0000;
  unique case (case_input_qualified)
    2'b00: case_output = 4'b0001;
    2'b01: case_output = 4'b0010;
    2'b10: case_output = 4'b0100;
    2'b11: case_output = 4'b1000;
    default: ;
  endcase
end
```

Though this can lead to timing issues. In the case in the linked Ibex issue the valid signal (which is whether or not the instruction is illegal) is fairly late signal which we don't want to factor into the decoder outputs that lead to the X going into a case statement input.

Another option is to allow the valid signal to qualify the assertion:

```verilog
`ASSERT(CaseInputKnown, case_valid |-> !$isunknown(case_input), clk_i, !rst_ni)
always_comb begin
  case_output = 4'b0000;
  unique case (case_input)
    2'b00: case_output = 4'b0001;
    2'b01: case_output = 4'b0010;
    2'b10: case_output = 4'b0100;
    2'b11: case_output = 4'b1000;
    default: ;
  endcase
end
```

I did also wonder if we could use assertions on the case output instead:

```verilog
`ASSERT_KNOWN(CaseOutputKnown, case_output, clk_i, !rst_ni)
[...]

Issue #13, opened 2019-12-09
Dealing with removing X in cases where consuming rather than producing X is the error


GregAC:
The current proposals for removing the use of X from RTL (https://github.com/lowRISC/style-guides/pull/12) focusses on cases where the generation of X would have been the error, so replacing the use of X with appropriate assertions and coding style changes is a sufficient replacement.

There are also areas where the X generation itself isn't because of an exceptional condition but rather flags a particular wire should have it's value ignored. So using the generated X value is the error you want to trap.

A typical example of this is on a bus interface. A valid signal accompanies some request information (address, type etc). When the valid signal is low the request information shouldn't be used and can be driven to X to signify this.

An example of a bug this can find is as follows (this could be a code fragment from a device acting on requests from the bus, NEVER_X is a macro that gives you an assert that fires if the specified signal becomes X)

```verilog
...
input logic         req_valid_i
input logic [7:0] req_addr_i
...

assign action_addr = req_addr_i == `ADDR_ACTION;

// Bug here, should have & req_valid_i
assign action_en = action_addr_i;

`NEVER_X(action_en);
```

When req_addr_i is driven to X when req_valid_i is low the assertion will quickly fire and the bug found an fixed. Where X is not used and the RTL doesn't tend to generate (ADDR_ACTION) as a value for req_addr_i when req_valid_i is low.

We can have explicit X driving in the DV environment on specific buses to help recreate this but you'd still want the NEVER_X macro or similar to ensure you get a quick, easy to diagnose failure. Plus you may want similar checking on internal signals that aren't well defined interfaces the DV can probe/drive values on (e.g. in Ibex we could set the write address of the register file to X when we're not writing a register).

One replacement would be for ``` `NEVER_X(action_en)``` is be replaced by an assertion ```action_en -> req_valid_i```. In the style guide we would recommend any 'enabling' signal that triggers an action has an assertion that checks if it is set the appropriate 'upstream enabling' signal is also set.

You can flip the above example on its head and instead have a bug where req_valid_i is being asserted when there is no valid request only this time we've fixed the bug above.

```verilog
assign action_en = action_addr_i & req_valid_i;
```

[...]

PR #27, opened 2020-02-17
Add further advice around Xs and assertion use

GregAC:
See #13 and #16 for discussion.

Signed-off-by: Greg Chadwick <gac@lowrisc.org>


VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,62 @@ reduction are significant enough to outweigh the possible synthesis mismatch
...
+
+One use case for `X` is where a signal is driven to `X` when it is invalid and
+shouldn't be consumed. For example on a bus interface the address lines may be
+driven to `X` when a valid transaction isn't present. Anything that incorrectly
+uses that invalid address will itself produce an `X` value that will be caught
+by an appropriate `` `ASSERT_KNOWN `` and trigger a simple to diagnose failure.
+
+For any value that could be invalid but won't be driven to `X` (as they are
+internal signals rather than generated by a DV environment) and is used to
+trigger some action (e.g. it factors into a enable for a register write) it is
```
sjgitty: *an* enable

VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,62 @@ reduction are significant enough to outweigh the possible synthesis mismatch
...
+
+#### Catching errors where invalid values are consumed
+
+One use case for `X` is where a signal is driven to `X` when it is invalid and
+shouldn't be consumed. For example on a bus interface the address lines may be
+driven to `X` when a valid transaction isn't present. Anything that incorrectly
+uses that invalid address will itself produce an `X` value that will be caught
+by an appropriate `` `ASSERT_KNOWN `` and trigger a simple to diagnose failure.
+
+For any value that could be invalid but won't be driven to `X` (as they are
```
sjgitty: subject of sentence is singular, so "as it is an internal signal"

VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,62 @@ reduction are significant enough to outweigh the possible synthesis mismatch
...
+// be X but must be ignored if reg_wr_en == 0
+assign reg_addr = ...
+assign reg_wr_en = ...
+
+...
+
+// trigger some specific action when a certain register is written
+logic special_reg_en;
+
+assign special_reg_en = (reg_addr == SPECIAL_REG_ADDR) && reg_wr_en;
```
sjgitty: the current VSG guidance is to use `&` here (still, though in discussion)

VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,62 @@ reduction are significant enough to outweigh the possible synthesis mismatch
...
+```systemverilog
+module mymod (
+  input [7:0] external_addr_i,
+  input       external_wr_en_i
+);
+
+  logic special_action_en;
+
+  assign special_action_en =
+    (external_addr_i == SPECIAL_ADDR) && external_wr_en_i;
```
sjgitty: ditto

VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,62 @@ reduction are significant enough to outweigh the possible synthesis mismatch
...
+driven to `X` when a valid transaction isn't present. Anything that incorrectly
+uses that invalid address will itself produce an `X` value that will be caught
+by an appropriate `` `ASSERT_KNOWN `` and trigger a simple to diagnose failure.
+
+For any value that could be invalid but won't be driven to `X` (as they are
+internal signals rather than generated by a DV environment) and is used to
+trigger some action (e.g. it factors into a enable for a register write) it is
+recommended an assertion is added to check that if the action is triggered then
+the value is valid. This triggers a similar simple to diagnose failure when an
+invalid value has been accidentally used.
```
sjgitty: The first sentence is very clause-heavy. Perhaps a simpler way to write it:

For an internally-generated signal that could be invalid (but not driven to `X`)
and is used to enable another signal (such as a register write-enable), it is
recommended to add an assert to check that when the enable is true, the signal
is valid. This triggers ...

GregAC:
Thanks for the review @sjgitty. It was never my intention to imply that non DV code could drive `X` to indicate invalid values. The first paragraph was intended as an intro discussing something you might like to do, but can't because of the the `X` limitations so suggests an alternative. However on reflection I think it's best not to overexplain in the style guide so I've just removed the first paragraph. Comparing the first and second code fragments should get the point across the first paragraph was aiming to make anyway.

VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,55 @@ reduction are significant enough to outweigh the possible synthesis mismatch
...
+#### Catching errors where invalid values are consumed
+
+For an internally-generated signal that could be invalid (but not driven to `X`)
+and is used to trigger some action (such as a register write-enable), it is
+recommened to add an assert to check that when the enable is true, the signal is
+valid. This triggers a simple to diagnose failure when an invalid value has been
+accidentally used.
+
+```systemverilog
+
```
tomroberts-lowrisc: I can see what this first example is driving-at having read the X-related issues, but I had to read it a couple of times to get there.

In a way this boils down to the recommendation "write assertions to check that the design isn't wrong" (and the example here is an assertion checking that an AND gate functions as an AND gate).

I would maybe just leave it out, but I appreciate that it then isn't addressing all the issues in #13 .

Otherwise LGTM.

GregAC: Yeah this particular example is a bit basic, but I do think such obviously true assertions do have value as when someone fiddles with the logic that makes it obviously true and messes it up you have a quick failure rather than a bug that may lurk for a while longer.

Maybe I'm just worrying about problems that don't really occur in practice!

VerilogCodingStyle.md
```diff
@@ -1808,6 +1808,55 @@ reduction are significant enough to outweigh the possible synthesis mismatch
 issues that the use of `X` literals may entail (especially with the gate-counts
 available in today's technologies).
 
+
+#### Catching errors where invalid values are consumed
+
+For an internally-generated signal that could be invalid (but not driven to `X`)
+and is used to trigger some action (such as a register write-enable), it is
+recommened to add an assert to check that when the enable is true, the signal is
```
msfschaffner: typo: recommended

VerilogCodingStyle.md
msfschaffner: Indeed, it spells out more specific guidance based the general recommendation above. But given the specific nature of the problem where X could be consumed, it is worth noting, hence LGTM as well.

PR #29, opened 2020-02-25
Add guidance on functions and tasks

### LINKED ISSUE #28: Use of functions and automatic
I don't think we currently have any guidance on the use of functions, in particular I think this is important for synthesizable  RTL. DV code may want different guidelines or be happy to have none. I'm purely considering synthesizable RTL here.

A few questions to answer:

1. Do we allow use of functions at all in synthesisable RTL?
2. Do we allow use of local variables within functions?
3. Do we mandate use of 'automatic' in all function definitions

I think our answers should be, yes we allow functions, that can use local variables and must always be declared automatic. We probably need some details around how local variables can be used within functions

Clearly functions are useful so saying no functions at all seems very limiting. Allowing local variables though can open up problematic behaviour, e.g.

```systemverilog
function [3:0] do_a_thing(input [3:0] foo);
  logic [3:0] state = 0;
  
  if (foo == ConstantValue) begin
    state = state + 1;
    return state;
  end else begin
    return foo;
  end
endfunction
```

By default `state` is static so this function has persistent state between calls. This has clear issues and shouldn't be allowed.

You can use the 'automatic' qualifier on a function to remove the default static state-holding behaviour I suspect there's some further gotchas around local variables in functions we need to be wary of.

GregAC:
Fixes #28

I've gone beyond the scope of what we've discussed in #28 in terms of rules/recommendations here. I am happy for these to be challenged just things that came up whilst I was writing it that I felt needing addressing. A brief list of what I've added on top if the #28 discussion is:

* No tasks, functions only
* Functions in packages or inside modules (not just outside a module in a non package .sv)
* Recommendations around use of `return result` vs `function_name = result`
* Specifying all local variables must get assigned  on all code paths

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -103,6 +103,7 @@ representation of this style guide.
     - [Generate Constructs](#generate-constructs)
     - [Signed Arithmetic](#signed-arithmetic)
     - [Number Formatting](#number-formatting)
+    - [Functions and Tasks](#functions-and-tasks)
     - [Problematic Language Features and Constructs](#problematic-language-features-and-constructs)
       - [Floating begin-end blocks](#floating-begin-end-blocks)
   - [Design Conventions](#design-conventions)
@@ -2289,6 +2290,156 @@ always_comb begin
     addr2 = 40'h41_c000_0000;
   end
 end
+```
+
+### Functions and Tasks
+
+The following section applies to synthesizable RTL only. See the [Coding Style
+Guide for Design Verification](DVCodingStyle.md) for DV usage.
+
+***In synthesizable RTL the use of functions is allowed, provided they are declared
+`automatic`. Tasks should not be used.***
+
+Functions must be declared in either a package or inside a module. A package is
+appropriate where the function relates to other definitions in the package and
+could be useful to multiple modules (even if it's currently only used by one). A
+module is appropriate where the function specifically relates to the internals
+of that module.
+
+Functions should aim to conceptually represent a reusable block of combinational
+logic.
+
+Storage types must be explicitly declared for all arguments and the function
+return value. All types must be 4-state data types, either `logic` or types
+derived from `logic` (such as appropriate `struct`, `enum` or `typedef` types).
+
+Do not use `output`, `inout`, or `ref` on function arguments. All functions
+should only consume inputs and produce one output. `input` is the default and is
+not required on the function arguments.
+
+
+&#x1f44e;
+```systemverilog {.bad}
+// - Doesn't have explicit storage type on `a` or `b` or return type
+// - `b` being used as `output` argument
+// - `input` not required on `a`
+function automatic [2:0] foo(input [2:0] a, [2:0] b);
+  b = b + 1;
+  return a + b;
+endfunction
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+// - Doesn't have explicit storage type on `a`, `b` or `c`
+// - Uses `output` on `c`
+// - `input` not required on `a` and `b`
+function automatic logic [2:0] foo(input [2:0] a, input [2:0] b, output [2:0] c);
+  c = a - b;
+  return a + b;
+endfunction
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+// - Uses 2-state data type `int` for `a`
+function automatic logic [2:0] foo(int a, logic [2:0] b);
+  return a + b;
+endfunction
+```
+
+&#x1f44d;
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+ return a ^ b;
+endfunction
+```
+
+&#x1f44d;
+```systemverilog {.good}
+typedef logic [2:0] bar_t;
+
+typedef struct packed {
+  logic [2:0] field;
+} baz_t;
+
+function automatic logic [2:0] foo(bar_t a, baz_t b);
+  return a + b.field;
+endfunction
+```
+
+Data should be returned from a function using an explicit `return result` style.
+Do not use a `function_name = result` style.
+
+&#x1f44e;
+```systemverilog {.bad}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  if (a == 3'd2) begin
+    foo = b;
+  end else begin
+    foo = a ^ b;
+  end
+endfunction
+```
+
+&#x1f44d;
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] result;
+
+  if (a == 3'd2) begin
+    result = b;
+  end else begin
+    result = a ^ b;
+  end
+
+  return result;
+endfunction
+```
+
+All local variables must be assigned in all code paths, either through an
+initial assignment or through the use of `else` and `default:` for `if` and
+`case` statements.
+
+&#x1f44d;
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] local_var_1;
+  logic [2:0] local_var_2;
+
+  local_var_1 = 3'd0;
+
+  if (a == 0) begin
+    local_var_1 = 3'd2;
+  end
+
+  unique case(b)
+    3'd0:    local_var_2 = 3'd1;
+    3'd1:    local_var_2 = 3'd3;
+    default: local_var_2 = 3'd0;
+  endcase
+
+  return local_var_1 + local_var_2;
+endfunction
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] local_var_1;
+  logic [2:0] local_var_2;
+
+  if (a == 0) begin
+    local_var_1 = 3'd2;
+  end
+
+  unique case(b)
+    3'd0:    local_var_2 = 3'd1;
+    3'd1:    local_var_2 = 3'd3;
+  endcase
+
+  return local_var_1 + local_var_2;
+endfunction
 ```
 
 ### Problematic Language Features and Constructs


VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
 ```
 
+### Functions and Tasks
+
+The following section applies to synthesizable RTL only.
+
+***In synthesizable RTL use of functions is allowed, provided they are declared
+automatic. Tasks should not be used.***
+
+Functions must be declared outside of a class in either a package or inside a
```
GregAC: I am now wondering if you can declare a function outside of a module/package? i.e

```systemverilog

function automatic logic [2:0] foo(logic [2:0] a);
...
endfunction

module bar();
...
endmodule
```

I think you can but if not this paragraph is redundant.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+&#x1f44e;
+```systemverilog {.bad}
+function automatic [2:0] foo(input [2:0] a, [2:0] b);
+  b = b + 1;
+  return a + b;
+endfunction
+```
+
+It is recommended that data is returned from a function using an explicit
+`return` style. Avoid using a `function_name = result` style other than for
```
GregAC: I had initially said `return` style only but noticed that the aes block used `function_name = result` in some cases.

My main reason for disliking that style is it can somewhat obscure the control flow that leads you to the result, especially if you have code paths that give you two or more `function_name = result` statements.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+```
+
+For any local variables ensure they will get assigned on all code-paths, whether
+through an initial assignment or through the use of `else` and `default:` for
+`if` and `case` statements
+
+&#x1f44d;
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] local_var_1 = 3'd0;
```
GregAC: Was wondering if this initial assignment along with definition is ok, couldn't immediately find anything in the LRM that suggests it would be problematic.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+
+&#x1f44e;
+```systemverilog {.bad}
+function automatic [2:0] foo(input [2:0] a, [2:0] b);
+  b = b + 1;
+  return a + b;
+endfunction
+```
+
+It is recommended that data is returned from a function using an explicit
```
GregAC: (Copying comment as I updated the file and it doesn't appear on the new version)

I had initially said return style only but noticed that the aes block used function_name = result in some cases.

My main reason for disliking that style is it can somewhat obscure the control flow that leads you to the result, especially if you have code paths that give you two or more function_name = result statements.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
 end
 ```
 
+### Functions and Tasks
+
+The following section applies to synthesizable RTL only.
```
GregAC: Once the DV style guide stuff is merged this should refer to the guidance for functions on DV. I'll await that PR to go in before this is merged and add a reference.

VerilogCodingStyle.md
sriyerg: I think it is understood that 'class' is a non-synthesizable construct, so we can omit it here. 

Functions declared in an SV file outside of a module or a package (yes, you can) is visible to the whole compilation unit AFTER that file is compiled. This is potentially adds unnecessary debug overhead. I think it should be limited to within the module / pkg / interface so that the scope is clear.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+
+***In synthesizable RTL use of functions is allowed, provided they are declared
+automatic. Tasks should not be used.***
+
+Functions must be declared outside of a class in either a package or inside a
+module. A package is appropriate where the function relates to other definitions
```
sriyerg: Wrap `package` & `module` in back-ticks since they are SV keywords?

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+in the package and could be useful to multiple modules (even if its currently
+only used by one). A module is appropriate where the function specifically
+relates to the internals of that module.
+
+Functions should aim to conceptually represent a reusable block of combinational
+logic.
+
+Storage types must be explicitly declared for all arguments and the function
```
sriyerg: Perhaps limit the args to only use 4-state data type `logic`?

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+return value.
+
+Do not use `input`, `output`, `inout`, or `ref` on function arguments.
```
sriyerg: Is there a reason to avoid these (except maybe `ref`)? The LRM doesn't indicate that these render the function unsynthesizable (AFAIK).

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+`if` and `case` statements
+
+&#x1f44d;
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] local_var_1 = 3'd0;
+  logic [2:0] local_var_2;
+
+  if (a == 0) begin
+    local_var = 3'd2;
```
sriyerg: I'm guessing this should be local_var_1 (ditto on 2386)?

VerilogCodingStyle.md
msfschaffner: I feel like `ref` should be excluded, and `inout` may be a candidate we may want to exclude because of style reasons (avoid tristate constructs on-chip). But for more than just one return value, `output` could be useful...

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+
+Functions should aim to conceptually represent a reusable block of combinational
+logic.
+
+Storage types must be explicitly declared for all arguments and the function
+return value.
+
+Do not use `input`, `output`, `inout`, or `ref` on function arguments.
+
+Do not assign to function arguments in the function body.
```
msfschaffner: This will be a function of whether we allow `output` above...

VerilogCodingStyle.md
msfschaffner: Although I agree that functions should be kept simple and just return one value if possible...

VerilogCodingStyle.md
msfschaffner: I think it _should_ be ok, lint is fine with it as well. In the past I always assigned defaults explicitly on a new line, though, to be on the safe side... maybe someone else has more experience with this variant?

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] local_var_1 = 3'd0;
+  logic [2:0] local_var_2;
+
+  if (a == 0) begin
+    local_var = 3'd2;
+  end
+
+  case(b)
```
msfschaffner: we strongly encourage to use `unique case` hence I would make this a unique case here as well...

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+```systemverilog {.bad}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+  logic [2:0] local_var_1;
+  logic [2:0] local_var_2;
+
+  if (a == 0) begin
+    local_var = 3'd2;
+  end
+
+  case(b)
```
msfschaffner: `unique case`

VerilogCodingStyle.md
imphil: It's different only for static functions (without `automatic`). 

To keep people from tripping over the differences, having the initialization in a separate line gives code that works the same in static and automatic functions. So I'd write:

```
logic [2:0] local_var_1;
local_var_1 = 3'd0;
```

VerilogCodingStyle.md
sjgitty: Just a quick note on this (will come back with
further review afterwards), it is legal to put a function
"floating" outside a module and class. Probably good
to leave this guidance here.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,109 @@ always_comb begin
...
+  b = b + 1;
+  return a + b;
+endfunction
+```
+
+It is recommended that data is returned from a function using an explicit
+`return result` style. Avoid using a `function_name = result` style other than
+for simple single-expression / no control flow functions.
+
+&#x1f44d;
```
sjgitty: I think this should be moved last after a correct example of your {.bad} version, like

&#x1f44e;
```systemverilog {.bad}
function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
  if (a == 3'd2) begin
    foo = b;
  end else begin
    foo = a ^ b;
  end
endfunction
```

&#x1f44d;
```systemverilog {.good}
function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
  logic [2:0] result;
  if (a == 3'd2) begin
    result = b;
  end else begin
    result = a ^ b;
  return result;
  end
endfunction

function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
 foo = a ^ b;
endfunction
```

VerilogCodingStyle.md
sjgitty: This is a style guide after all, so I think it is a good idea to limit it to good practice,
which IMO is all inputs and just the one output. So if we're going with this particular
guidance, I would restate it as follows, since `input` per se isn't a problem.

  Do not use `output`, `inout`, or `ref` on function arguments. All functions should
  only consume inputs and produce the one function output. `input` is the default and
  is not required on the function arguments.

Then probably add this to your {.bad} examples

&#x1f44e;
```systemverilog {.bad}
function automatic logic [2:0] foo(input [2:0] a, input [2:0] b, output [2:0] c);
  c = a - b;
  return a + b;
endfunction
```

VerilogCodingStyle.md
eunchan: Should the `result` be automatic?

imphil:
@GregAC Looks like this PR is almost there, can you incorporate the last suggestions from Scott?

GregAC:
Yup don't worry haven't forgotten about it.

I think the only real point of discussion (beyond corrections, small improvements/additions etc) was how restrictive we want to be on what functions can take as arguments.

I'd quite purposefully gone for a function as a stand-in for a block of combinational logic style and said input only via arguments with only one output.

Multiple outputs are useful but of course still possible, you just need to concatenate things or use a struct as a return type. In my experience people generally don't do super-complex things with functions in RTL so I don't think disallowing output arguments will cause major issues.

I'll update the style guide today assuming we're happy with that idea but give the other commenters a chance to reply to continue the discussion if they still disagree with that style.

VerilogCodingStyle.md
sjgitty: this has landed now if you want to add a reference.

VerilogCodingStyle.md
sjgitty: recommend resolving this conversation and leaving as is.

VerilogCodingStyle.md
sjgitty: +1

VerilogCodingStyle.md
sjgitty: +1

VerilogCodingStyle.md
sjgitty: recommend this modification.

VerilogCodingStyle.md
sjgitty: agree: redundant with above guidance, can be removed.

VerilogCodingStyle.md
sjgitty: recommend this movement, though i have an indentation error above (`end` should be
before `return result;`

VerilogCodingStyle.md
sjgitty: agreeing with @imphil

VerilogCodingStyle.md
sjgitty: agreeing with @sriyerg

VerilogCodingStyle.md
sjgitty: agreeing with @msfschaffner

VerilogCodingStyle.md
GregAC: I decided against this as there's plenty of other places within the style guide that refer to module and package without the back-ticks. I think it partially comes down to keyword as noun vs keyword as keyword:

> Foos can only be declared within a package

vs

> There must only be one usage of `package` within a file

I will have a scan through this PR and identify any keyword as keyword usages and add back-ticks if they don't already exist.

VerilogCodingStyle.md
GregAC: I think it's useful to be able to pass enums and structs etc so I'll expand that to 4-state data type `logic` and datatypes derived from `logic`.

Also I think it's useful to have functions that work on integer parameters (a classic example is log2 though we have an in-built function for that) but to get this PR landed I'll forgot about that for now and add an issue.

GregAC:
Apologies for the slow updates, have now incorporated all of the feedback, though a couple of places where I disagreed, see the comments.

Thanks for the reviews @sriyerg @sjgitty @imphil @msfschaffner

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
 end
 ```
 
+### Functions and Tasks
+
+The following section applies to synthesizable RTL only. See the [Coding Style
+Guide for Design Verification](DVCodingStyle.md) for DV usage.
+
+***In synthesizable RTL use of functions is allowed, provided they are declared
```
imphil: use -> the use

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+`automatic`. Tasks should not be used.***
+
+Functions must be declared in either a package or inside a module. A package is
+appropriate where the function relates to other definitions in the package and
+could be useful to multiple modules (even if its currently only used by one). A
```
imphil: its -> it's/it is

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+module is appropriate where the function specifically relates to the internals
+of that module.
+
+Functions should aim to conceptually represent a reusable block of combinational
+logic.
```
imphil: Maybe highlight *combinatorial*

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+
+Storage types must be explicitly declared for all arguments and the function
+return value. All types must be 4-state data types, either `logic` or types
+derived from `logic` (such as appropriate `struct`, `enum` or `typedef` types).
+
+Do not use `output`, `inout`, or `ref` on function arguments. All function
```
imphil:
```suggestion
Do not use `output`, `inout`, or `ref` on function arguments. All functions
```

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+should only consume inputs and produce one output. `input` is the default and is
+not required on the function arguments.
+
+
+&#x1f44e;
+```systemverilog {.bad}
+// - Doesn't have explicit storage type on 'a' or `b` or return type
```
imphil: Choose the same quoting style for a and b

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+// - `b` being used as `output` argument
+// - `input` not required on a
```
imphil: also quote a for consistency.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+function automatic [2:0] foo(input [2:0] a, [2:0] b);
+  b = b + 1;
+  return a + b;
+endfunction
+```
+
+&#x1f44e;
+```systemverilog {.bad}
+// - Doesn't have explicit storage type on 'a', `b` or `c`
```
imphil: as above.

VerilogCodingStyle.md
imphil: I see no good reason for the `function_name = result` style, so I'd completely disallow it and change the AES block.

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+```
+
+&#x1f44d;
+```systemverilog {.good}
+function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
+ foo = a ^ b;
+endfunction
+```
+
+For any local variables ensure they will get assigned on all code-paths, whether
```
imphil:
```suggestion
All local variables must be assigned in all code paths, either
```

VerilogCodingStyle.md
```diff
@@ -2291,6 +2292,165 @@ always_comb begin
...
+through an initial assignment or through the use of `else` and `default:` for
+`if` and `case` statements
```
imphil:
```suggestion
`if` and `case` statements.
```

VerilogCodingStyle.md
GregAC: @vogelpi (or anyone else) do you wish to offer any defense for this style or are you happy for it to be disallowed?

VerilogCodingStyle.md
GregAC: Considered it, thought it ended up looking a bit odd so left it.

eunchan:
I also like this approach. Thanks! LGTM

VerilogCodingStyle.md
vogelpi: I wouldn't mind at all to change those functions in the AES module! I think having an explicit return value is cleaner and much more readable.

GregAC: Cool in that case I will switch it to allowing `return result` only.

GregAC:
Hitting the merge button on this as I think everyone's more or less happy with it. If there is something people disagree with feel free to continue the discussion, we can always do another PR to change things.

Issue #24, opened 2020-02-11
[verilog stlye guidelines] parameters should be all caps


mwbranstad:
I have noticed that verilog passed parameters are recommended to be coded as camelcase style. I think this is bad idea. My only argument is that all capital letters is the defacto standard in the industry. Every style guideline that i have every read or created states this. 

It is clear that this style is ubiquitous. I notice that under the heading of Parameterized Types in the verilog style guidelines, there are examples showing all capitial letter parameters there. Even if this issue is rejected, some work needs to be done to clean up the style guidelines in its present state.

Additionally, I believe this applies to both local parameters and emul types.

msfschaffner:
related to https://github.com/lowRISC/ibex/issues/579#issuecomment-584379063

mwbranstad:
> Personally, I think using PascalCase (UpperCamelCase in the style guide) in the tunable parameter is good even though it is slightly off from the industrial standard. the borderline isn't that obvious between what the tunable parameter is and what constant is, but at least the different form of the word helps me to distinguish.
> 
I am trying to think of case where a parameter is not "tunable". I am guessing that constants are the case where you want to pass a non-tunable, but trying to figure out why one would do this. Anyway, thanks for the explanation and history.

sjgitty:
(Similar apology for slowness of response. Thanks @mwbranstad for bringing these up.)

Oh boy, this is a bit uglier than I thought. https://github.com/lowRISC/stwg-base/issues/427 raised an issue, solved it in https://github.com/lowRISC/stwg-base/pull/749, then left this note "This will next be rolled into the already-public lowRISC style guide". But that never happened. :scream:  (Sorry for the pre-history references).

This means that our still-dangling-for-no-good-reason `doc/rm/verilog_style_guide.md`
in lowrisc/opentitan is *newer than this one*! Not a good state of affairs.

Let's do this:
1. @mwbranstad read through the style guide in opentitan, see if it is better or worse than what is here w.r.t. your concerns. I don't think it "solves" them. Update this issue with the thoughts w.r.t. that.
2. @sjgitty will rectify the discrepancy of the two style guides ASAP with a PR here
3. @sjgitty will remove the opentitan VSG since we've been wanting to for a while.
4. We need an answer on https://github.com/lowRISC/opentitan/issues/988, which is how/if to render

My apologies for letting it get to this state of affairs.

imphil:
[...]

Issue #21, opened 2020-01-27
verilog style guidelines: Logical versus Bitwise


mwbranstad:
I have read this part of the verilog style guidelines:
Logical vs. Bitwise
Use logical constructs for logical comparisons, bit-wise for data.

Logical constructs (!, ||, &&, ==, !=) should be used for all constructs that are evaluating logic (true or false) values, such as if clauses and ternary assignments. Use bit-wise constructs (~, |, &, ^) for all data constructs, even if scalar.

I reread the style guidelines again on this. I concluded that the guidelines could be roughly rewritten like this: "For control path statements, use logical operators. For data path statements, use bit wise operators, even if the data is a single bit." 

So for control logic, these examples apply:
1) if (fifo_pop && enable[0]) then
2) assign fifo_push = (fifo_not_full && enable) ? fifo_write : 1'b0;
3) assign fifo_push = fifo_not_full && enable && fifo_write;

For data path:
1) assign write_bus[31:0] = raw_bus[31:0] & {32{write_valid}}; (vector case)
2) assign write_bus[0] = raw_bus[0] & write_valid; (scalar case)

This issue has been opened to clean up any ambiguities in the style guidelines.

imphil:
@mwbranstad Thanks for opening this issue. I've moved it to the repository that contains the style guide for further discussion there.

rasmus-madsen:
Thank for this suggestion @mwbranstad I like the suggested clean up.

cdgori:
I also like the proposed changes, they make sense intuitively to me.

vogelpi:
Hi @mwbranstad ,
thanks for opening this issue. I think your third example for control logic
```
assign fifo_push = fifo_not_full && enable && fifo_write;
```
is the only case in which your suggestion is different from the current guidelines which recommend
```
assign fifo_push = fifo_not_full & enable & fifo_write;
```

To me, using bitwise operators in this case seems more natural as you generate a 1-bit signal by combining multiple 1-bit signals. Of course, if you are describing the same functionality using the synthax as in your Example 1 or 2, the boolean operations are more natural as you formulate a boolean expression. We could probably argue for hours about this and personal preferences... But I think the really important point here is that - unfortunately this together with some reference papers got removed from the style guide in a previous PR - what logical/boolean operators do with `X` inputs.

[...]

PR #32, opened 2020-03-31
[doc] Three new optional styles added

mwbranstad:
There are three new additional styles added to the verilog style guide. They are the following:
- allowance of all caps for parameters
- allowance of boolean operators on control assignment statements
- allowance of an alternate format for module port list

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -966,12 +966,16 @@ facilitate the creation of IP that may be re-used across many projects.
 
 The preferred naming convention for all immutable constants is to use `ALL_CAPS`, but there are times when the use of `UpperCamelCase` might be considered more natural.
 
+The naming convention for parameter and localparam allows for the use of ALL_CAPS or UpperCamelCase.
+UpperCamelCase is the preferred style for mutable parameters, but there is allowance for ALL_CAPS to accommodate the common style found in other code bases.
+
+
 | Constant Type | Style Preference | Conversation |
 | ---- | ---- | ---- |
 | \`define            | `ALL_CAPS`       | Truly constant |
-| module parameter    | `UpperCamelCase` | truly modifiable by instantiation, not constant |
-| derived localparam  | `UpperCamelCase` | while not modified directly, still tracks module parameter |
-| tuneable localparam | `UpperCamelCase` | while not expected to change upon final RTL version, is used by designer to explore the design space conveniently |
+| module parameter    | `UpperCamelCase` or `ALL_CAPS` | truly modifiable by instantiation, not constant |
+| derived localparam  | `UpperCamelCase` or `ALL_CAPS` | while not modified directly, still tracks module parameter |
+| tuneable localparam | `UpperCamelCase` or `ALL_CAPS` | while not expected to change upon final RTL version, is used by designer to explore the design space conveniently |
 | true localparam constant | `ALL_CAPS`  | Example `localparam OP_JALR = 8'hA0;` |
 | enum member true constant | `ALL_CAPS` | Example `typedef enum ... { OP_JALR = 8'hA0;` |
 | enum set member | `ALL_CAPS` or `UpperCamelCase`     | Example `typedef enum ... { ST_IDLE, ST_FRAME_START, ST_DYN_INSTR_READ ...`, `typedef enum ... { StIdle, StFrameStart, StDynInstrRead...`. A collection of arbitrary values, could be either convention. |


mwbranstad:
Addresses issue #24

VerilogCodingStyle.md
```diff
@@ -965,12 +965,14 @@ facilitate the creation of IP that may be re-used across many projects.
 
 The preferred naming convention for all immutable constants is to use `ALL_CAPS`, but there are times when the use of `UpperCamelCase` might be considered more natural.
 
+The naming convention for parameter and localparam allows for the use of `ALL_CAPS` or `UpperCamelCase`. The allowance for `ALL_CAPS` is to accommodate this ubiquitous style found in the ASIC industry, and by all third party IP vendors.
```
sjgitty, line 968: I have no problem with allowances for `ALL_CAPS`, but I think our code base
is preferring `UpperCamelCase`, so that should be our leader. Plus you've
undermined our past decisions a bit with the wording. This sentence plus
the table below imply that `ALL_CAPS` is preferred, which isn't what we've
agreed upon. How about this as a suggestion:

The naming convention for `parameter` and `localparam` allows for the use of `ALL_CAPS` or `UpperCamelCase`.
`UpperCamelCase` is the preferred style for mutable parameters, but there is allowance for `ALL_CAPS` to accommodate the common style found in other code bases.

VerilogCodingStyle.md
```diff
@@ -965,12 +965,14 @@ facilitate the creation of IP that may be re-used across many projects.
...
+
 | Constant Type | Style Preference | Conversation |
 | ---- | ---- | ---- |
 | \`define            | `ALL_CAPS`       | Truly constant |
-| module parameter    | `UpperCamelCase` | truly modifiable by instantiation, not constant |
-| derived localparam  | `UpperCamelCase` | while not modified directly, still tracks module parameter |
-| tuneable localparam | `UpperCamelCase` | while not expected to change upon final RTL version, is used by designer to explore the design space conveniently |
+| module parameter    | `ALL_CAPS` or `UpperCamelCase` | truly modifiable by instantiation, not constant |
```
sjgitty, line 973: Please switch the order here since our code base is predominantly `UpperCamelCase`.

    `UpperCamelCase` or `ALL_CAPS`

VerilogCodingStyle.md
```diff
@@ -1390,10 +1392,14 @@ Use the Verilog-2001 combined port and I/O declaration style. Do not use the
 Verilog-95 list style. The port declaration in the module statement should fully
 declare the port name, type, and direction.
 
-The opening parenthesis should be on the same line as the module declaration,
-and the first port should be declared on the following line.
+The opening parenthesis may be on the same line as the module declaration,
```
sjgitty, line 1395: I'm with @msfschaffner on this one that it needs more discussion. We
can spawn to a different PR or leave this one open for a while to get other
opinions. @asb @imphil @eunchan @tjaychen @GregAC @martin-lueker 

I'm not an emacs user, so please weigh in if this is a true hardship. Even if
it is hard to avoid, I don't believe it should be in our rationale explicitly as
on the line below.

VerilogCodingStyle.md
```diff
@@ -2385,9 +2407,10 @@ a check for `X` on the 4-state net before resolving to a 2-state variable.
 ***Use logical constructs for logical comparisons, bit-wise for data.***
 
 Logical constructs (`!`, `||`, `&&`, `==`, `!=`) should be used for all
-constructs that are evaluating logic (true or false) values, such as
-if clauses and ternary assignments.  Use bit-wise constructs (`~`, `|`,
-`&`, `^`) for all data constructs, even if scalar.
+constructs that are evaluating control path logic. This includes boolean
+logic (true or false) values, such as if clauses and ternary assignments.
```
sjgitty, line 2411: I have a recommended rewrite for this section in #21, let's do that in a
separate PR please. I can initiate that one. I feel this wording goes
all the way to the other side of the current guidance, while #21 was
recommending allowing both styles in different contexts.

VerilogCodingStyle.md
rasmus-madsen: please include me if this spawns into another issue
@sjgitty  with emacs default is to align following lines to the opening parenthese
so an example could be

```
module this_is_my_mode (
                        input logic a,
                        input logic b,
                        output logic c
                         );
```

vs

```
module this_is_my_mode
   (
    input logic a,
    input logic b,
    output logic c
   ); 

```
this can of cause be edited in the mode file - but then one would have to change this every time going to or from opentitan work.

imphil: @rswarbrick as you're a heavy Emacs user: can we provide a project-specific config for OpenTitan to make it easier for Emacs user to get the right style from verilog-mode?

rswarbrick: As far as I can tell, this isn't trivial to do in [verilog-mode](https://github.com/veripool/verilog-mode) (the major mode that everyone uses for editing Verilog in Emacs). When calculating how to do indentation, it currently decides that the ports are in a syntactic context that it calls `cparenexp` (parenthesized expression context), and always lines up with the opening paren. However, it's maintained by Wilson Snyder (of Verilator fame), who's super helpful if you send him a pull request.

This has been annoying me too, so I intend to get around to it soonish (hah..). I'd be thrilled if someone else did so first! :-) In the meantime, rectangular text commands and judicious use of  `indent-rigidly` (C-x TAB) seems to be a reasonable workaround.

I'm not sure what Rasmus means by "mode file". If we have a working way to setup verilog mode in the "lowRISC style", the usual thing would be to write a function like this:

    (defun setup-lowrisc-verilog-style ()
      (when (lowrisc-repo-p (buffer-file-name))
        (setq-local my-variable 'some-value)))

and install it on `verilog-mode-hook`. This runs when you open a Verilog file (well, when you enable verilog-mode, but that should be equivalent) and sets up "lowRISC mode" if needed.

Obviously, you also have to write an implementation of `lowrisc-repo-p`. I'd probably do something like walk upwards through the path looking for the first directory with a `.git` subdirectory and then checking whether that directory name was in a specified list (`ibex`, `opentitan`, ...). Alternatively, we could use a dir-locals file checked in to the repo which set an "I'm a lowRISC repository" flag.

If there are significantly more people that would use this than me and Rasmus, we could even have a simple "emacs setup" repository somewhere.

VerilogCodingStyle.md
vogelpi: I totally agree with @sjgitty here. `UpperCamelCase` should be clearly marked as preferred style.

vogelpi: As for allowing the use of `ALL_CAPS` for (local) parameters: I am against it. Once people start to use both, it won't be any more distinguishable if something is constant or tunable during code review. And this was the whole point of the original discussion. But this is my personal opinion. I would like to hear other peoples voices.

VerilogCodingStyle.md
```diff
@@ -2385,9 +2407,10 @@ a check for `X` on the 4-state net before resolving to a 2-state variable.
 ***Use logical constructs for logical comparisons, bit-wise for data.***
 
 Logical constructs (`!`, `||`, `&&`, `==`, `!=`) should be used for all
-constructs that are evaluating logic (true or false) values, such as
-if clauses and ternary assignments.  Use bit-wise constructs (`~`, `|`,
-`&`, `^`) for all data constructs, even if scalar.
+constructs that are evaluating control path logic. This includes boolean
```
vogelpi, line 2410: I think the suggestion was that:
1) Logical operators should only be allowed outside if clauses and ternary assignments if they describe control path logic, and
2) A comment should be added saying that this statement is indeed describing control path logic (to clarify the designer's intent).

This was proposed here: https://github.com/lowRISC/style-guides/issues/21#issuecomment-598493498

mwbranstad:
Reduced scope to only the all caps topic. Used @sjgitty phrasing in update.

mwbranstad:
After re-reviewing this PR, I have to admit that I agree with @vogelpi, in that adding an all caps option dilutes the original style. I am going to retract this PR.
However, if I was around when the original discussion occurred, I would have pushed for all caps only, but that window has passed.

The style guide is generally quite good. I think a common style that promotes productivity and easier development is a must. The *_q and *_d notation is a long time favorite, and the module *_i and *_o is a recent favorite. I think the fsm guidelines are better than what I have used.

I still do not believe where the parenthesis sits on a module statement does not measure up to the above standard. The code is not better understood one way or another.

For boolean operators, I thought I was following the original intent with my example. Will leave this up to someone else to make it clearer.

mwbranstad:
Hi Scott,

I still have this item open. I asked for three changes originally, but have given up on the first two (all caps parms, module format). I would like to know if I can code my control statements with Boolean operators. I think the issue is that I use assignment statements for control instead of if-then-else style.

I though you were going to propose new language to explain this. Please let me know what path you would like to take here.

BTW, I have noticed a number of “prim” modules that do not follow some of the guidelines. For example, the use of *_d and *_q for flop inputs and outputs is not used, a style I have used for many years.

msfschaffner:
@mwbranstad, I am pretty sure that we still have code in our repo that does not fully adhere to the style guide. so if you come across something that you think is inconsistent you could either file an issue for that file, or if it is a quick and small fix, just make a PR for it. I am also cleaning up such things when I come across them...

mwbranstad:
@msfschaffner ok thanks for the guidance.

sjgitty:
> Hi Scott, I still have this item open. I would like to know if I can code my control statements with Boolean operators. I think the issue is that I use assignment statements for control instead of if-then-else style. I though you were going to propose new language to explain this. Please let me know what path you would like to take here.
#38 should address this @mwbranstad . Sorry for the delays.

> BTW, I have noticed a number of “prim” modules that do not follow some of the guidelines. For example, the use of *_d and *_q for flop inputs and outputs is not used, a style I have used for many years.
Agreed. If you're feeling energetic, you can create a PR, otherwise feel free to file an issue and we'll pull it off the pile eventually. The issue should be in the opentitan repo.

PR #34, opened 2020-04-09
[dv style guide] More macro usage guidelines

sriyerg:
This was prompted by
[this PR](https://github.com/lowRISC/opentitan/pull/1918)
made on OpenTitan, which was seemingly hard to debug.

### DIFF
--- a/DVCodingStyle.md
+++ b/DVCodingStyle.md
@@ -22,6 +22,7 @@ following the [UVM methodology](https://www.accellera.org/images//downloads/stan
     * [Guidelines for UVM library macros](#guidelines-for-uvm-library-macros)
     * [Guidelines for macros within dv_macros.svh](#guidelines-for-macros-within-dv_macrossvh)
     * [General macro use guidelines](#general-macro-use-guidelines)
+    * [Typical gotchas with macro usage](#typical-gotchas-with-macro-usage)
   * [Factory](#factory)
   * [Configuration Mechanism](#configuration-mechanism)
   * [Sequences](#sequences)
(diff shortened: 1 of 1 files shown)


DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+
+    assign x = `AND(p || q, r) || s;
+    // Correct: ((p || q) && (r)) || s
+    ```
+2.  Macro usage scope and avoiding namespace collisions.
+
+    *   If macros are defined for use in multiple sources, place them in a
+        separate SystemVerilog header file with appropriate `ifdef` guards.
+
+        This header file must not be compilation unit (provided as an input to
```
rswarbrick, line 497: Nit: "must not be *a* compilation unit"

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
         the end of the file, otherwise it will pollute the global namespace.
 
+#### Typical gotchas with macro usage:
+
+While it is better to avoid macros altogether, they do make our lives easy by
+shortening pieces of repeated code, in a manner that cannot be achieved with a
+function or a task. In general, the following describes the best practices to
+avoid common gotchas with macro usage.
+
+1.  Always wrap macro arguments **and** the output as applicable, within
```
rswarbrick, line 472: I think this would be clearer as something like this: "Always wrap macro arguments in parentheses. If the macro evaluates to an expression, also wrap the whole expansion in parentheses."

Otherwise, the reader has to work out what "as applicable" means.

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+    parenthesis.
+
+    The following example illustrates how not wrapping the formal macro
+    arguments in parenthesis can cause unexpected behavior.
```
rswarbrick, line 476: Nit: parentheses

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+        tools do not even warn about namespace collisions, causing unnecessary
+        debug overhead. It is a good practice to `undef` macros at the end of
+        the file so that they are no longer visible to all subsequent
+        compilation units.
+3.  Avoid passing expressions as formal macro arguments.
+
+    As illustrated in the example below, it is better to avoid passing
+    expressions as arguments, since they are replaced verbatim. The macros may
+    reference the input arguments in multiple places, casuing the expressions to
+    be evaluated multiple times, resulting in unforeseen side effects.
```
rswarbrick, line 530: This only applies to side-effecting expressions. Since defining local variables is a bit of a bore in Verilog, let's not disallow things like `` `FOO(a+b)``.

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+
+    :-1:
+    ```systemverilog
+    `define CHECK_AND_UPDATE(a, b) if (a != b) a = b;
```
rswarbrick, line 534: This definition is missing parentheses around `a` and `b` (see previous style guide item :-D)

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+    ```systemverilog
+    `define UPDATE_VALUES(a, b) \
+      begin                     \
+        a = 88;                 \
+        b = 1955;               \
+      end
+
+    if (power == 1.21) `UPDATE_VALUES(speed, year)
+    ```
+5.  Never invoke macros on the same line as the conditonal statement it is invoked
```
rswarbrick, line 571: Nit: conditional.

Also, is this actually necessary? If people are following the previous instruction, this makes no difference.

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+    within. Always use `begin` and `end` when doing so.
+
+    This is a supplimental solution to avoid the previous gotcha.
```
rswarbrick, line 574: Nit: supplemental

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+    // different and unexpected values.
+    ```
+    :+1:
+    ```systemverilog
+    `define CHECK_AND_UPDATE(a, b) if (a != b) a = b;
+
+    y++;
+    `CHECK_AND_UPDATE(x, y)
+    ```
+4.  Wrap multi-line macros in `begin` and `end`.
```
rswarbrick, line 547: I just re-read the [OpenTitan PR #1918](https://github.com/lowRISC/opentitan/pull/1918) that seems to have prompted this and I don't think this guidance is enough. The problem with the `` `downcast` macro there is that, while it is a single statement, it has an `if` at top-level, so behaves weirdly if followed by an `else`.

This is the problem that C developers usually address with the `do { ... } while (0)` trick. I think the right guidance here is probably something like "Make sure that macros that evaluate to one or more statements behave just like a single statement, wrapping with `begin/end` if necessary". Then give an example of a macro with an `if` statement, together with what can go wrong.

GregAC:
Have yet to read this in detail but we may want these in the main style guide as we use macros in non-DV RTL and they will have the same gotchas. Perhaps we want the bulk of this in VerilogCodingStyle.md with anything DV specific in DVCodingStyle.md?

asb:
> Have yet to read this in detail but we may want these in the main style guide as we use macros in non-DV RTL and they will have the same gotchas. Perhaps we want the bulk of this in VerilogCodingStyle.md with anything DV specific in DVCodingStyle.md?

+1. It feels like the majority of this is non-DV specific and so would be best added to VerilogCodingStyle.md

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
         If the macro is defined in a local perspective, it must be undefined at
         the end of the file, otherwise it will pollute the global namespace.
 
+#### Typical gotchas with macro usage:
+
+While it is better to avoid macros altogether, they do make our lives easy by
```
sjgitty, line 467: ^easy^easier^

DVCodingStyle.md
sjgitty: The point of `(`parenthesizing`)` the arguments is under the assumption that there will be expressions passed as arguments, so I think we need to allow for this.

sjgitty:
@sriyerg I think this is close, but there are a few requested changes here.

sjgitty:
> @sriyerg I think this is close, but there are a few requested changes here.

@sriyerg this is lingering for a long time. Is it still relevant? If so, can you finish it off? If not, can you amend or close?

DVCodingStyle.md
rswarbrick: Reading this again (prompted by Scott's prod), I think you don't want to talk about multi-line macros at all. The number of lines is not the point; it's the number of statements. Also (as I said above), an `if` statement is especially pesky and needs the begin/end treatment.

If you want a blanket rule, you could say "wrap non-expression macro bodies in `begin/end`", but I'm not keen. I think that it's reasonable to expect a macro author to understand the language, so I prefer what I suggested in the comment above.

rswarbrick: It turns out that Wilson Snyder has an excellent list of SystemVerilog macro gotchas called [Verilog Preprocessor: Force for \`Good and \`Evil](https://www.veripool.org/papers/Preproc_Good_Evil_SNUGBos10_paper.pdf). On page 5 (under the "Semicolon Eater" heading) he talks about this `if` statement problem and proposes a `do begin ... end while (0)`, just like the usual C solution.

sriyerg:
Wow, sorry about this. I will try to close it by this weekend.

DVCodingStyle.md
```diff
@@ -461,6 +462,125 @@ utility macros for even more common code sequences.
...
+    // Correct: ((p || q) && (r)) || s
+    ```
+2.  Macro usage scope and avoiding namespace collisions.
+
+    *   If macros are defined for use in multiple sources, place them in a
+        separate SystemVerilog header file with appropriate `ifdef` guards.
+
+        This header file must not be compilation unit (provided as an input to
+        the tool for compilation). Instead, it should be `` `included`` in the
+        package, module or the interface that intends to invoke them.
```
sriyerg, line 499: Need to rewrite this section - `` `included`` files get compiled, so the "must not be in a compilation unit" statement is invalid.

DVCodingStyle.md
sriyerg: Done.

DVCodingStyle.md
sriyerg: Done.

DVCodingStyle.md
sriyerg: Removed.

DVCodingStyle.md
sriyerg: I have reworded this now. 
Yup - see the end of the section I am already referencing that same paper :)

DVCodingStyle.md
sriyerg: Yeah, removed.

DVCodingStyle.md
sriyerg: Removed.

sriyerg (commented):
I am closing this PR due to some issue. Please see #45 instead - I have addressed your comments.

PR #36, opened 2020-04-24
Clarify the usage of `parameter` in packages

vogelpi:
This PR clarifies that within packages `parameter` should be used instead of `localparam`.

This PR resolves lowRISC/style-guides#35.


imphil:
Thanks Pirmin!

Issue #30, opened 2020-03-05
[UVM:styleguide] best end of test practices 


rasmus-madsen:
from #22 a discussion on where and how to use objections has been started.
to not clutter that PR too much I have taken discussion this into its own issue.

The original suggestion is to avoid objections in drivers and monitors and instead add them in the scoreboard.

I will try to make an argument why monitors is the perfect place for objections and why they should only go into phase_ready_to_end() 

one of the Arguments of why it is nice to have objections in the scoreboard is given by @sriyerg 
I took the liberty of quoting it here:

> I see some valid usecases in allowing raising/dropping objections in the scoreboard; example: in xbar type DUT, raise obj when transaction enters the DUT and drop only when it exits. This objection (that helps ensure that the test does not exit prematurely) cannot be captured in the monitor and has to be captured in a higher level component. It cannot be captured in the test either since it only generates and sends C-R stimulus to the driver. It has no knowledge of when the transactions actually enter and exit the DUT.
> 
> We keep the performance impact low by raising objection exactly once (dont raise if it is already raised), which is what the line implies. You can find that logic in our base scoreboard (`hw/dv/sv/dv_lib/dv_base_scoreboard.sv`). If by deadlocks you mean hangs, then there is an obvious problem either in the design or dv, that needs fixing. We avoid hanging forever through reasonable test timeouts.

I have two arguments against this - the first one being simply performance.
every time you raise and drop an objection is causing the component hierarchy to be traversed.
so for shallow environments this might be a minor impact but as the hierarchy gets deeper this quickly gets expensive.

The other argument is that implementing this in the scoreboard requires a watchdog timer to be implemented a long the side to avoid deadlocks.

Here is what I suggest

Most modules have deterministic input to output latency and does not need any objections besides test case objections and a good choice of drain time. this should be the default go to.
objections should only be used outside the test when the DUT latency is not detministic!

Some modules does not have deterministic latency. an example can be some sort of packet builder where input packets are grouped in larger chunks and these chunks are only forwarded when reaching a certain size.

Modules like this usually have an internal timeout to flush out undersized chunks.

In such a scenario it will be hard to raise objections in a scoreboard as the scoreboard cannot easily predict the number of outputs based on the number of inputs.
and the suggested objection method would be insufficient.

instead I suggest putting the objetion inside the phase_ready_to_end() function.
this is called once all modules have dropped their run_phase objections and the test is trying to end the run phase.

here is what is currently in my scoreboard (AES) 

```
[...]

PR #37, opened 2020-05-04
[dv/style] Handling EOT with phase_ready_to_end

udinator:
This PR attempts to encapsulate the discussion around end-of-test practices (namely using `phase_ready_to_end()`) in #30 and add it to the DV style guide.

The code example of `phase_ready_to_end()` is taken from lowrisc/opentitan#2137.

This PR also contains some very minor changes to get rid of the word `you`  which is used in several locations.

Signed-off-by: Udi Jonnalagadda <udij@google.com>

### DIFF
--- a/DVCodingStyle.md
+++ b/DVCodingStyle.md
@@ -247,8 +247,8 @@ having to recompile and allows testlists to be created for regression runs.
     uvm_field_array_int`, etc. within `uvm_object/component_utils_begin` and
     `uvm_object/component_utils_end`.
 2.  For all field automation macros, use `UVM_DEFAULT` as the flag argument.
-    For example, if you want to enable everything except comparing and printing for
-    a `rx_delay` field, the macro would look like:
+    For example, if it is desired that everything except comparing and printing
+    be enabled for some `rx_delay` field, the macro would look like this:
 
     ```systemverilog
     `uvm_field_int(rx_delay, UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
@@ -546,7 +546,7 @@ When using the UVM configuration database, these guidelines should be followed:
     locations where they are needed. It should not be used to pass integers,
     strings, or other basic data types, since it is much easier for namespace
     collisions to occur when using lower level types.
-5.  You must always check for success when accessing the `uvm_config_db`.
+5.  Checks for success must always be performed when using the `uvm_config_db`.
     If the lookup fails, issue a `uvm_fatal` to terminate the simulation.
     This should be done as follows:
 
@@ -594,30 +594,105 @@ cause race conditions if there are any parallel threads.
 
 ### Objections and Coordinating End of Test
 
-1.  At a high level, a phase should not end when either there is more stimulus
-    to send during that phase or the DUT has yet to respond to some stimulus that
-    has been sent.
-    To prevent a phase from ending in the former case, raise and lower objections
-    inside tests.
-    To prevent a phase from ending in the latter case, raise objections in the
-    scoreboard whenever expect queues are not empty and lower them when they are empty.
-2.  It is recommended to avoid using objections in drivers or monitors
-    for performance reasons they should be used inside these components
-    sparingly and only when needed.
-    (Note: this point is still under discussion, and is still TBD)
-3.  If a sequence needs to raise/drop objections, it should call
-    `raise_objection` in its `pre_start()` method and `drop_objection` in its
-    `post_start()` method. A good alternative to this is to call
-    `set_automatic_phase_objection(1)` in the sequence constructor.
-4.  Always perform the test `if (get_starting_phase() != null)` before calling
-    `raise_objection` or `drop_objection` within a sequence.
-5.  When starting a sequence that can raise and drop objections, if you want
-    the sequence to raise and drop objections, set the `starting_phase` member of
-    the sequence object the `<sequence_handle>.set_starting_phase(<uvm_phase>)`
-    before starting the sequence.
-6.  To avoid reimplementing objection functionality in multiple sequence classes,
-    create a parent sequence that implements this functionality for its derivatives.
-7.  When calling `raise_objection` or `drop_objection`, you must pass a string
+
+As a general principle, a phase should not end when either there is more stimulus to
+send during that phase or the DUT has yet to respond to some stimulus that has
+been sent.
+
+To prevent a phase from ending when there is more stimulus to send, a test class
+must raise and lower objections.
+
+To prevent a phase from ending before the DUT has responded to some stimulus,
+classes derived from `uvm_component` should raise and lower objections  in their
+`phases_ready_to_end()` method.
+
+It is also recommended to only include objections in the monitor component.
+When using this approach, these guidelines should be followed:
+
+1.  The monitor's `phase_ready_to_end()` method should implement a watchdog timer
+    that raises an objection if any traffic is seen on the bus and lowers it once
+    no traffic is seen within the timeout period. If traffic is seen, the
+    watchdog resets its timer.
+    Refer to the example below:
+
+    ```systemverilog
+    // Example code for the monitor class.
+    // This code assumes that the associated configuration object has an integer
+    // property "ok_to_end_delay_ns", which gives the window timeout in
+    // nanoseconds.
+
+    protected bit ok_to_end = 1'b1;
+    protected bit watchdog_done = 1'b0;
+
+    function void phase_ready_to_end(uvm_phase phase);
+     if (phase.is(uvm_run_phase::get())) begin
+       if (watchdog_done) fork
+         monitor_ready_to_end();
+       join_none
+       if (!ok_to_end || !watchdog_done) begin
+         phase.raise_objection(this, $sformatf("%s objection raised", `gfn));
+         fork
+           begin
+             // wait until ok_to_end is set
+             watchdog_ok_to_end();
+             phase.drop_objection(this, $sformatf("%s, objection dropped",
+             `gfn));
+           end
+         join_none
+       end
+     end
+
+    endfunction
+
+    // This watchdog waits for ok_to_end_delay_ns nanoseconds while checking for
+    // any traffic seen on the bus during this period.
+    // If traffic is seen, the watchdog will restart until it sees no traffic.
+    task watchdog_ok_to_end();
+      fork
+        begin : isolation_fork
+          bit watchdog_reset;
+          fork
+            forever begin
+              // check bus interface for traffic
+              @(ok_to_end or watchdog_reset);
+              if (!ok_to_end && !watchdog_reset) watchdog_reset = 1;
+            end
+            forever begin
+              #(cfg.ok_to_end_delay_ns * 1ns);
+              if (!watchdog_reset) begin
+                break;
+              end else begin
+                watchdog_reset = 0;
+              end
+            end
+          join_any;
+          disable fork;
+
+          watchdog_done  = 1;
+        end : isolation_fork
+      join
+    endtask
+
+    // This task is invoked as a non-blocking thread when phase_ready_to_end()
+    // is first entered.
+    //
+    // This task should monitor the DUT's interface and control ok_to_end: it
+    // should be cleared whenever there are any pending transactions and
+    // set to 1 if there are no pending transactions.
+    virtual task monitor_ready_to_end();
+    endtask
+    ```
+
+2.  It is recommended not to use other objections in the scoreboard; the
+    scoreboard should just make use of information already available to it
+    (like fifo size, etc...) to determine when it is ready to end, as the
+    monitor has already guaranteed that no more traffic is seen on the bus.
+    These fifo checks should be checked during its `check_phase()`.
+
+The more general guidelines below should be followed when dealing with
+end-of-test timeouts or objections:
+
+1.  When calling `raise_objection` or `drop_objection`, a string must be passed
     as a second argument to describe the objection to help with debugging.
 
     ```systemverilog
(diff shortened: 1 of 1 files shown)


DVCodingStyle.md
```diff
@@ -594,30 +594,122 @@ cause race conditions if there are any parallel threads.
...
+
+At a high level, a phase should not end when either there is more stimulus to
+send during that phase or the DUT has yet to respond to some stimulus that has
+been sent.
+
+To prevent a phase from ending in the former case, raise and lower objections
+inside the test class.
+
+To prevent a phase from ending in the latter case, it is required to raise and
+lower objections inside the `phase_ready_to_end()` function.
```
rswarbrick, line 606: I think this should be "method", not "function". And probably this needs combining with the following paragraph (so that you can say whose method).

DVCodingStyle.md
```diff
@@ -594,30 +594,122 @@ cause race conditions if there are any parallel threads.
...
+It is also recommended to only include objections in the monitor component.
+When using this approach, these guidelines should be followed:
+
+1.  The monitor's `phase_ready_to_end()` function should implement a watchdog timer
```
rswarbrick, line 610: s/function/method

DVCodingStyle.md
```diff
@@ -594,30 +594,122 @@ cause race conditions if there are any parallel threads.
...
+    that raises an objection if any traffic is seen on the bus and lowers it once
+    no traffic is seen within the timeout period; if traffic is seen, the
+    watchdog resets its timer.
+    Refer to the example below:
+
+    ```systemverilog
+    function void phase_ready_to_end(uvm_phase phase);
```
rswarbrick, line 617: In general, this is an enormous code snippet for a style guide. Would it be better to point at existing code in (say) OpenTitan as an example?

DVCodingStyle.md
```diff
@@ -594,30 +594,122 @@ cause race conditions if there are any parallel threads.
...
+     uvm_phase current_phase;
+     string str;
+     uvm_object objectors[$];
+
+     current_phase = phase;
```
rswarbrick, line 622: If this snippet is staying in: What does taking a reference to `phase` do here?

DVCodingStyle.md
```diff
@@ -626,14 +718,14 @@ cause race conditions if there are any parallel threads.
...
+2.  It is recommended to use the built-in UVM timeout mechanism,
+    `uvm_root::set_timeout(<time_limit>, 1)`.
     The base test in every testbench should specify a default timeout limit,
     which can be overridden by every derivative test.
     A standard plusarg `+UVM_TIMEOUT=...` is recommended to be used for this
     override mechanism.
-    Alternatively, some separate timeout mechanism can be created and utilized
-    within the testbench.
+    Alternatively, some separate global timeout mechanism can be created and
+    utilized within the testbench.
```
rswarbrick, line 728: Do we have examples of tests that use a different timeout mechanism? If not, maybe we should drop this exception - we already mandate that UVM tests derive from the `dv_base_...` infrastructure that provides the suggested version, no?

DVCodingStyle.md
rasmus-madsen: I agree this is a big snippet. but I still like in the style guide as pointers to code need to be maintained.

DVCodingStyle.md
rasmus-madsen: @rswarbrick I think this was taken directly from the example I provided.
I don't recall exactly why I did that.

it seem that one could just just phase instead.

DVCodingStyle.md
rasmus-madsen: I agree with @rswarbrick unless we already have code in place that deviates from this suggestion we should enforce the use of the build in timeout

DVCodingStyle.md
udinator: You're probably right, I'll change this to 'method' instead of 'function'.

RE your second point - the general intent was to specify two separate recommendations, 1) that all objections should only be touched inside of `phase_ready_to_end()` and 2) we generally recommend objections should only be touched in the monitor.

DVCodingStyle.md
udinator: I tend to agree with @rasmus-madsen here, I agree that this is a large code snippet, but I also do prefer that it remains in the style guide rather than point to some other code files elsewhere in the OT repository; additionally due to the relative complexity of this watchdog mechanism, I feel having a "bare-bones" example here is more clear than pointing to an already-implemented monitor that might introduce some extra complexity into its implementation.

rswarbrick: Ok. Is there anything that can be cut out of this example? I mean, "It's really easy to do this, just write the following 50 lines of code" doesn't make our house style sound great :-)

DVCodingStyle.md
udinator: good point, I missed this when copying over the example code - I'll delete this.

DVCodingStyle.md
udinator: Hm, there's some code that gets and prints out any objects that still have raised objections that may not be the most necessary for this snippet (starting with the call to `phase.get_objection().get_objectors(objectors)`), and I did add numerous comments to explain what's going on.
@rasmus-madsen wdyt?

DVCodingStyle.md
udinator: That's true, I believe the existing approach is done using `::set_timeout(...)`, I'll delete the second part.

udinator:
The force push deletes the section about alternate timeout mechanisms, and changes the word `function` to `method` when referring to `phase_ready_to_end()`.

DVCodingStyle.md
weicaiyang: I simplified the logic a bit, perhaps we can discuss it here https://github.com/lowRISC/opentitan/pull/2137, and then see what codes should be put here.

rswarbrick: That looks good to me. Do we suggest that code derives from `dv_base_...` classes? If so, maybe this style guide should say something like "derive from `dv_base_monitor`; poke `ok_to_end` as follows..." and we don't need to write out the implementation here.

rasmus-madsen:
LGTM I'm ok with you merging

DVCodingStyle.md
udinator: lowRISC/opentitan#2137 looks pretty concise in its implementation - I will update this example code here to match that structure as well.

DVCodingStyle.md
```diff
@@ -594,30 +594,120 @@ cause race conditions if there are any parallel threads.
...
+             wait_for_ok_to_end();
+             phase.drop_objection(this, $sformatf("%s objection dropped", `gfn);
+             // Once this point has been reached, it is a good idea to check if
+             // any other objects are still objecting.
+             phase.get_objection().get_objectors(objectors);
+             foreach(objectors[i]) begin
+               // Print information about any objects that
+               // still have raised objections.
+               ...
+             end
```
sriyerg, line 650: I think you can leave it at the suggestion above.

DVCodingStyle.md
```diff
@@ -594,30 +594,122 @@ cause race conditions if there are any parallel threads.
...
+             // previously raised objection can now be safely dropped.
+             wait_for_ok_to_end();
+             phase.drop_objection(this, $sformatf("%s objection dropped", `gfn);
+             // Once this point has been reached, it is a good idea to check if
+             // any other objects are still objecting.
+             current_phase.get_objection().get_objectors(objectors);
+             foreach(objectors[i]) begin
+               // Print information about any objects that
+               // still have raised objections.
+               ...
```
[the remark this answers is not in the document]
udinator: Will do.

DVCodingStyle.md
```diff
@@ -594,30 +594,120 @@ cause race conditions if there are any parallel threads.
...
+      end
+    endtask
+
+    // This watchdog will wait for timeout_delay cycles while checking for any
+    // traffic on the bus during this period.
+    // If traffic is seen before timeout_delay cycles, the watchdog will keep
+    // repeat this process until traffic has stopped.
+    task watchdog_ok_to_end();
+      bit reset_timer;
+      fork
```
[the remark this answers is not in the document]
udinator: ah good point

DVCodingStyle.md
udinator: Yup, replaced all of this with a comment.

DVCodingStyle.md
udinator: @rswarbrick w.r.t. derivation from the `dv_base` and `cip_lib` libraries, the style guide remains agnostic in this regard - it tries as much as possible not to make any assumptions about any DV library dependencies (apart from `dv_macros.svh` and `prim_assert.sv`).

DVCodingStyle.md
```diff
@@ -594,30 +594,101 @@ cause race conditions if there are any parallel threads.
...
-    the sequence to raise and drop objections, set the `starting_phase` member of
-    the sequence object the `<sequence_handle>.set_starting_phase(<uvm_phase>)`
-    before starting the sequence.
-6.  To avoid reimplementing objection functionality in multiple sequence classes,
-    create a parent sequence that implements this functionality for its derivatives.
-7.  When calling `raise_objection` or `drop_objection`, you must pass a string
+
+At a high level, a phase should not end when either there is more stimulus to
+send during that phase or the DUT has yet to respond to some stimulus that has
+been sent.
```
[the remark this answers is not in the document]
udinator: Thanks for pointing this out - reworded to "As a general principle".

DVCodingStyle.md
```diff
@@ -594,30 +594,101 @@ cause race conditions if there are any parallel threads.
...
+
+To prevent a phase from ending in the former case, it is required to raise and
+lower objections inside the test class.
```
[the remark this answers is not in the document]
udinator: Updated, thanks

DVCodingStyle.md
```diff
@@ -594,30 +594,101 @@ cause race conditions if there are any parallel threads.
...
+
+To prevent a phase from ending in the latter case, it is required to raise and
+lower objections inside the `phase_ready_to_end()` method of classes derived
+from `uvm_component`.
```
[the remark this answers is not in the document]
udinator: Updated

DVCodingStyle.md
```diff
@@ -594,30 +594,101 @@ cause race conditions if there are any parallel threads.
...
+It is also recommended to only include objections in the monitor component.
+When using this approach, these guidelines should be followed:
+
+1.  The monitor's `phase_ready_to_end()` method should implement a watchdog timer
+    that raises an objection if any traffic is seen on the bus and lowers it once
+    no traffic is seen within the timeout period; if traffic is seen, the
+    watchdog resets its timer.
```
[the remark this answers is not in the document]
udinator: Updated, thanks

DVCodingStyle.md
```diff
@@ -594,30 +594,101 @@ cause race conditions if there are any parallel threads.
...
+    Refer to the example below:
+
+    ```systemverilog
+    // This example code assumes existence of three  members:
+    // 1) class variable `protected bit ok_to_end`.
+    // 2) class variable `protected bit watchdog_done`.
+    // 3) class variable `int ok_to_end_delay_ns` in the configuration object.
```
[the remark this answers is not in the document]
udinator: Hmm, yes I also think just writing the declarations would be cleaner also.
Indeed it does seem like this code example sort of half-assumes existence of some mysterious monitor base class that this is derived from, hopefully this change (and the ones below) should rectify this.

DVCodingStyle.md
```diff
@@ -594,30 +594,101 @@ cause race conditions if there are any parallel threads.
...
+        end : isolation_fork
+      join
+    endtask
+
+    // This task is invoked as a non-blocking thread when phase_ready_to_end()
+    // is first entered.
+    //
+    // This task must be overridden and implemented in the IP-level testbench
+    // to set the ok_to_end bit to prevent simulations from finishing if there
+    // are any pending transactions in the monitor.
```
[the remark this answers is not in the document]
udinator: Yep agreed, good catch.

DVCodingStyle.md
```diff
@@ -803,9 +872,9 @@ subtle errors.
 
 ### Code Within Asserts
 
-For any randomization that is required, you must use the macros specified
-in the [Randomization Macros section](#macro-usage), do not use `assert()` to check
-manual randomization calls, use the provided macros instead.
+For any randomization that is required, the macros specified in the
+[Randomization Macros section](#macro-usage) must be used. Do not use `assert()`
+to check manual randomization calls, use the provided macros instead.
```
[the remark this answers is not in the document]
udinator: Sure, I'll add some description to the commit message to clarify that some minor unrelated changes are included in this PR.

udinator:
> Hi Udi! Thanks for doing this, and I realise that it's taken quite a while. I've got a few more-or-less nitty comments below, which I think should probably be fixed before this goes in.
> 
> The general thing that I think needs fixing (addressed by many of the specific comments below) is that we've got a sort of "half-dv_lib" recommendation at the moment: the text is sort of explaining how you might rewrite dv_lib and assumes that you'll have a monitor base class etc. I don't think that's quite right.
> 
> If we want this to be free-standing, and not talk about our DV libraries, I think the right thing is to say "A monitor should do XYZ" and leave it to the reader to work out how the class hierarchy works. Maybe we should add a note somewhere in the text saying that dv_lib exists and that you get some of this functionality for free by using it?

I agree, looking back at this there seems to be some wording that implies there is some parent monitor class that everything inherits from. I've gone through and changed the wording of the comments to make it more explicitly 'free-standing'. Thanks for the comments!

udinator:
@weicaiyang would you mind taking a look at this just to make sure it aligns with the code in your OT PR?

udinator:
Thanks @weicaiyang !

PR #40, opened 2020-07-18
[vsg] add guidance on closing braces

### LINKED ISSUE #39: [SV] Placement of closing parentheses in initializer lists
Our style guide currently has this example:

```systemverilog
assign structure = '{
    src: src,
    dest: dest,
    default: '0};
```

Note the placement of the closing `};`.

In the wild, we have a fair bit of code which uses the following style:

```systemverilog
assign structure = '{
    src: src,
    dest: dest,
    default: '0
};
```

(e.g. https://github.com/lowRISC/opentitan/blob/a4a9e4013612607c68086ce4534a14206928f3a6/hw/ip/prim/rtl/prim_keccak.sv#L191-L198)

Currently our style guide doesn't have language (beyond that example) mandating one use over the other. Which option do we prefer?

sjgitty:
Signed-off-by: Scott Johnson <scottdj@google.com>

Resolves #39 

Add clarity on termination characters.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -406,22 +406,26 @@ four spaces, like this:
 ```systemverilog {.good}
 assign zulu = enabled && (
     alpha < bravo &&
-    charlie < delta);
+    charlie < delta
+);
 
 assign addr = addr_gen_function_with_many_params(
     thing, other_thing, long_parameter_name, x, y,
-    extra_param1, extra_param2);
+    extra_param1, extra_param2
+);
 
 assign structure = '{
     src: src,
     dest: dest,
-    default: '0};
+    default: '0
+};
 ```
 
 Or, if it improves readability, align the continued part of the expression with
 a grouping open parenthesis or brace, like this:
 
-```systemverilog
+:+1:
+```systemverilog {.good}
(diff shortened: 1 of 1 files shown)


VerilogCodingStyle.md
```diff
@@ -437,6 +441,26 @@ assign structure = '{src: src,
 Operators in a wrapped expression can be placed at either the end or the
 beginning of each line, but this must be done consistently within a file.
 
+Open syntax characters such as `{` or `(` that end one line of a multi-line
+expression should be terminated with close characters (`}`, `)`) on their
+own line. Examples:
```
rswarbrick, line 446: Does this need something explicit to allow the `zulu = ... delta` syntax from above?

sjgitty: > Does this need something explicit to allow the `zulu = ... delta` syntax from above?

Do you mean line 429? That is theoretically explained by the "improves human readability" portion.

rswarbrick: I'm sorry, I was being slow. Yes, I did mean that line, and I thought this paragraph was disallowing that syntax. But the point is that this says "... `(` that end one line...", which doesn't apply to
```
enabled && (foo ||
            bar)
```

So all is well!

PR #45, opened 2020-10-09
[dv style guide] More macro usage guidelines

sriyerg:
This was prompted by
[this PR](https://github.com/lowRISC/opentitan/pull/1918)
made on OpenTitan, which was seemingly hard to debug.

### DIFF
--- a/DVCodingStyle.md
+++ b/DVCodingStyle.md
@@ -22,6 +22,7 @@ following the [UVM methodology](https://www.accellera.org/images//downloads/stan
     * [Guidelines for UVM library macros](#guidelines-for-uvm-library-macros)
     * [Guidelines for macros within dv_macros.svh](#guidelines-for-macros-within-dv_macrossvh)
     * [General macro use guidelines](#general-macro-use-guidelines)
+    * [Typical gotchas with macro usage](#typical-gotchas-with-macro-usage)
   * [Factory](#factory)
   * [Configuration Mechanism](#configuration-mechanism)
   * [Sequences](#sequences)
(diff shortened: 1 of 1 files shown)


DVCodingStyle.md
```diff
@@ -454,13 +455,194 @@ utility macros for even more common code sequences.
...
+        ```
+
+        Generally, there is no need to `` `include`` this header file if it
+        can be guaranteed that it is already compiled prior to the current
+        compilation unit in which those macros are invoked. This is typically
+        handled by an external tool that works out the dependencies across
+        the compilation units (example: FuseSoC). However, it may not always be
+        possible to guarantee this, and the `` `include`` may be necessary.
+        The `` `ifndef`` guard helps prevent the re-definition of the macros with
+        itself.
```
rswarbrick, line 559: I agree that this statement is true, but I'm not sure it's good style. I'm in favour of trying to remember to include all headers that you use, and relying on include guards to avoid duplication. Otherwise, you get weird problems when you change the includes in one file, and a completely different one suddenly breaks.

DVCodingStyle.md
```diff
@@ -454,13 +455,194 @@ utility macros for even more common code sequences.
...
+        // packages sources
+        `include "foo_env_cfg.sv"
+        ...
+        `include "foo_scoreboard.sv"
+        `include "foo_env.sv"
+
+
+        `undef GET_FOO_TX_FIFO_DATA
+        // Good: All package sources can invoke this macro. It will no longer
+        //       be visible beyond this compilation unit.
```
rswarbrick, line 614: Does SystemVerilog have a notion of "compilation unit" like C and similar languages? Surely, the whole pain of undefining macros is because it doesn't?

Also, I think it might be clearer to read this if the two halves of the sentence were joined. So something like this, maybe? "All package sources can invoke this macro, but it is not visible in any other code".

DVCodingStyle.md
```diff
@@ -454,13 +455,194 @@ utility macros for even more common code sequences.
...
+        ```
+
+        :+1:
+        ```systemverilog
+        // file: foo_env_pkg.sv
+        `define GET_FOO_TX_FIFO_DATA(baz) ...
+
+        ...
+
+        // packages sources
```
rswarbrick, line 605: Nit: I think this should probably by "package sources"

DVCodingStyle.md
```diff
@@ -454,13 +455,194 @@ utility macros for even more common code sequences.
...
+        `include "foo_scoreboard.sv"
+        `include "foo_env.sv"
+
+
+        `undef GET_FOO_TX_FIFO_DATA
+        // Good: All package sources can invoke this macro. It will no longer
+        //       be visible beyond this compilation unit.
+        ```
+
+3.  Wrap macros that resolve to multiple statements within `begin` and `end`.
```
rswarbrick, line 617: The paper you cite actually suggests a `do begin ... end while (0)` loop, which also avoids the ugly problem of needing semicolons after some macros, but not others. Is there a reason not to include that?

DVCodingStyle.md
```diff
@@ -454,13 +455,194 @@ utility macros for even more common code sequences.
...
+    this bug by wrapping the macro definition itself in `begin`..`end`.
+
+    :-1:
+    ```systemverilog
+    `define UPDATE_VALUES(a, b) \
+      a = 88;                   \
+      b = 1955;
+
+    if (power == 1.21) `UPDATE_VALUES(speed, year)
+    // Bad: The year is unconditionally set to 1955.
```
rswarbrick, line 630: Great Scott! I love this example! (Although isn't there a factor of 10^9 missing somewhere?) :-D

DVCodingStyle.md
sriyerg: Fixed.

DVCodingStyle.md
sriyerg: Reworded a bit.

DVCodingStyle.md
sriyerg: Fixed :)

DVCodingStyle.md
sriyerg: Wrapping the macro defs inside `do begin` ... `end while (0)` will require you to put a semi-colon at the end of a block of code. The following are the reasons for not mandating it:

- A block of code does not need semi-colon at the end. SV allows us to put arbitrary `begin..end` sections unlike C where arbitrary `{...}` would be a syntax error. 
- It makes it easy to distinguish macros that expand to a block of code (no semi-colon) vs macros that expand to an expression (semi-colon added). The ugliness you mention actually serves as an identification.
- This is in-line with UVM, which also does not require us to put semi-colon at the end of macros that expand into a block of code.
- Our OT codebase uses this style all over the place, so if we mandate `do begin ... end while (0)` it will be hard to update our codebase to match this style.  
- If we explicitly say  `do begin ... end while (0)` is also allowed, the 2nd bullet will no longer be true, and that will make the problem even uglier.

DVCodingStyle.md
rswarbrick: I'm not sure this matches how we currently use files like `prim_assert.svh`. Also, note that this style guide doesn't actually mention FuseSoc, so I don't think it makes sense to talk about how FuseSoc works (which I suspect might be tool-dependent anyway) here.

@imphil: I know you've fixed a couple of missing includes like this. What do you think?

DVCodingStyle.md
rswarbrick: Hmm, ok. I think I probably prefer semicolons everywhere but I guess UVM has made that decision for us already.

DVCodingStyle.md
sriyerg: Whatever build system you use to manage your SV sources, it will ultimately generate a dependency - resolved file list (the .f file is kind of an industry standard that all ASIC tool flows use) that is fed to the ASIC tool. FuseSoC is provided as an example - all ASIC build systems will work the same way.

imphil: The big caveat is "if it can be guaranteed that it is already compiled prior to the current compilation unit". We neither require nor guarantee nor implicitly fulfill this behavior at the moment for all EDA tools.

Background: Most EDA tools can be used in a way that generates a single compilation unit for the whole project, or in a way that generates a single compilation unit per .sv source file (i.e. how C/C++ compilation is typically done). (And typically arbitrary in-between approaches.)
Edalize, at the moment, does not guarantee either behavior. Even more, the different backends implement things differently: the VCS and Vivado backends, for example, put everything into a single compilation unit. The Aldec Riviera-PRO backend, on the other hand, creates individual compilation units.


Personal comment: I prefer including all dependencies in each file. This is a common approach in most programming languages, all preferring explicit coding over implicit behavior. I'd therefore remove the whole statement from the style guide, and possibly even turn it around to require all dependencies to be included.

sriyerg: @imphil, regardless if whether you provide one monolithic compilation unit to the simulator, or individual one for each package, unlike C, macros definitions once compiled, are visible to all sources / compilation units without needing to `` `include`` header files repeatedly. This is because in SV, a header file is yet another source like any other SV source.

Like I said in my previous comment - if you add the header source as a dependency in the FuseSoC core file, that should itself guarantee that the macro header is compiled first. I see the problem with `prim_assert` - in its core file, it has the tag **`is_include_file: true`**, which does not place it in the file list for compilation immediately, but only adds its  search path (`+incdir+<path>`). This forces the user to not only `` `include`` it in all sources, but also list it as a FuseSoC core dependency (so that the search path is added to all "compilation units"). It also forces the addition of `ifndef` guards to prevent its own recompilation! So I'm really not sure how is the extra line `` `include prim_assert.svh`` added to every single source even useful. If you remove the `is_include_file: true` tag, all you need is the FuseSoC core dependency (which IMO is already explicit enough). 

In any case, the current text already recommends adding the `` `ifndef`` protection against macro-redefinitions with itself due to the header `` `included`` multiple times. I will reword it a bit. 

Considering UVM macros are used ALL over the place without actually including `uvm_macros.svh`, I will not be adding a recommendation to explicitly include the macro header files in every single source.

sriyerg: I have removed the controversial line altogether. PTAL, thanks!

rswarbrick (approved):
This looks fine to me. I wonder whether the text should actually be in the design style guide (since DV code isn't the only thing to use macros), but let's leave that for a future commit.

sriyerg:
> This looks fine to me. I wonder whether the text should actually be in the design style guide (since DV code isn't the only thing to use macros), but let's leave that for a future commit.

Use of macros should be avoided in the design. At the moment, it is only limited to SVA specific ones.

PR #46, opened 2020-11-10
DV style guide: mention non-compliant usage of disable

imphil:
Only `disable fork` is a valid way to terminate all sub-processes;
`disable fork_process_label` works on VCS (and probably other
simulators), but isn't LRM-compliant, and not supported/silently ignored
by Xcelium.

See also https://github.com/lowRISC/ibex/issues/1174.

### DIFF
--- a/DVCodingStyle.md
+++ b/DVCodingStyle.md
@@ -1074,8 +1074,30 @@ assert(function_that_has_side_effects());
 Always put `wait fork` and `disable fork` constructs inside of an
 `isolation fork...join` block to avoid erroneous waiting.
 It is recommended to use the `DV_SPINWAIT` macro whenever possible,
-[as described here](#macro-usage)
+[as described here](#macro-usage).
 
+When disabling subprocesses, this may be done by using `disable fork` to
+terminate all processes or by `disable thread_label` to disable a specific
+thread. You may not use `disable fork_label` to disable all threads in a fork
+statement because there is inconsistent support for this among EDA tools. The
+use of `disable fork_label` is not compliant to the SystemVerilog-2017
+standard (see Section 9.6.2 and 9.6.3 of the SV2017 LRM). Example code:
+
+```
+fork: fork_label
+  begin: thread_label
+  end
+  ...
+join_any
+
+// INCORRECT.
+disable fork_label;
+// Valid, but does not respect parent child process relationships.
+disable thread_label;
+// Kills all threads spawned by the parent thread, including the ones in the
+// past.
+disable fork;
+```
 
 ### Void Casts
 


imphil:
@PaulOKCrevinn @CathalMCrevinn does this language sound about right?

DVCodingStyle.md
```diff
@@ -1074,7 +1074,10 @@ assert(function_that_has_side_effects());
 Always put `wait fork` and `disable fork` constructs inside of an
 `isolation fork...join` block to avoid erroneous waiting.
 It is recommended to use the `DV_SPINWAIT` macro whenever possible,
-[as described here](#macro-usage)
+[as described here](#macro-usage).
```
sriyerg, line 1077: nit: Seems like this file is using the 80 char per line 'README' file convention.

DVCodingStyle.md
```diff
@@ -1074,7 +1074,10 @@ assert(function_that_has_side_effects());
...
+
+Only `disable fork` may be used to terminate all subprocesses of the calling processes;
```
sriyerg, line 1079: This is incorrect. `disable label` is a perfectly valid way to stop the execution of the code block. I have explained in the Ibex bug what is actually wrong in that code. 

```systemverilog
  fork: label
    begin: thread1
    end
    ...
   join_any
  disable label; // INCORRECT
  disable thread1; // Perfectly valid, but does not respect parent child process relationships
  disable fork; // Kills all threads spawned by the parent thread, including the ones in the past
```

Depending on the situation, `disable thread1` and `disable fork`, both can be used. The above section clarifies the pitfall of `disable fork` and the need for an isolation fork. If `thread1` does not spawn more child threads, then `disable thread1` will obviate the need for the isolation fork which is required with `disable fork`. 

Please update this to say `disable label` is invalid only if `label` is the label given to the `fork` itself rather than the child threads. Also, indicate the pitfall of `disable thread1`.

sriyerg (changes requested):

DVCodingStyle.md
imphil: Sri, could you please suggest a specific wording to add to avoid going back and forth on this? Thanks! (Note that this text is in the "`"Wait and Fork" section.)

imphil:
@sriyerg with the Verible check for this functionality now close to landing, could you please propose s specific wording for the style guide?

DVCodingStyle.md
marnovandermaas: Should be fixed now.

DVCodingStyle.md
marnovandermaas: I've proposed a solution here, which hopefully complies.

marnovandermaas (approved):
I'm going to approve this PR as this has been dormant for a while. Feel free to submit a follow-up PR if you would like to change this wording.

Issue #48, opened 2021-04-15
How to align named ports in module instantiations?


imphil:
How do we want to format named port lists in module instantiations in SystemVerilog source code? The style guide is currently silent on this topic, but uses in examples the "align" formatting shown below.

Verible give us the following options:

```
    --named_port_alignment (Format named port connections:
      {align,flush-left,preserve,infer}); default: infer;
```

`align` and `flush-left` are the actual choices. `preserve` doesn't change formatting, and `infer` tries to guess the formatting based on what the code used, which isn't a good choice for a style guide.

## Option 1: align
Verible with `--named_port_alignment=align`. Note that tabular alignment is always done in "blocks" (as indicated by an empty line in between).

```systemverilog
module my_module;

  mod u_mod (
      .clk_i,
      .rst_ni,
      .in_1 (my_signal_in),
      .out_2(my_signal_out),

      .in_another_block_i(my_signal_in),
      .in3               (something)
  );

endmodule
```

## Option 2: flush-left
Verible with `--named_port_alignment=flush-left`:

```systemverilog
module my_module;

  mod u_mod (
      .clk_i,
      .rst_ni,
      .in_1(my_signal_in),
      .out_2(my_signal_out),

      .in_another_block_i(my_signal_in),
      .in3(something)
  );

endmodule
```

sriyerg:
Option 1 preferred.

msfschaffner:
I would also have a slight preference for option 1.

rswarbrick:
Option 1 for me. Is it possible to add an extra space (so there is a gap after `out_2` and `in_another_block_i`)?

mdhayter:
Option 1 seems closest to our current style.

Do comments restart the indentation or only completely blank lines? It seems useful to be able to put comments at the top of groups of signals without changing the indentation?

GregAC:
Personally I like alignment (so Option 1), I find it makes it easier to understand connectivity at a glance. I am not too fussed about the precise details of that alignment.

imphil:
@mdhayter wrote:
> Do comments restart the indentation or only completely blank lines? It seems useful to be able to put comments at the top of groups of signals without changing the indentation?

Only at least one empty line indicates a new block, as the following example shows. This allows comments within a block and is consistent with e.g. Markdown, Latex or other languages which use an empty line as block/paragraph separator.

@rswarbrick wrote:
> Is it possible to add an extra space (so there is a gap after out_2 and in_another_block_i)?

[...]

PR #50, opened 2021-04-19
Format module port expressions in tabular style

imphil:
Port expressions in module instantiations should be formatted in tabular style, as discussed in #48.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -459,7 +459,7 @@ assign bus_concatenation = {
 };
 
 inst_type inst_name1 (
-  .clk_i    (clk),
+  .clk_i       (clk),
   .data_valid_i(data_valid),
   .data_value_i(data_value),
   .data_ready_o(data_ready)
@@ -521,21 +521,47 @@ mymodule mymodule(.a(a),.b(b));
 
 #### Tabular Alignment
 
-***Adding whitespace to cause related things to align is encouraged.***
+Tabular alignment groups two or more similar lines so that the identical parts are directly above one another.
+This alignment makes it easy to see which characters are the same and which characters are different between lines.
 
-Where it is reasonable to do so, align a group of two or more similar lines so
-that the identical parts are directly above one another. This alignment makes it
-easy to see which characters are the same and which characters are different
-between lines.
+***The use of tabular alignment is generally encouraged.***
+
+***The use of tabular alignment is required for some constructs as detailed in the corresponding subsection of this guide.***
+
+Constructs which require tabular alignment:
+
+* [Port expressions in module instantiations](#module-instantiation)
+
+Each block of code, separated by an empty line, is treated as separate "table".
 
 Use spaces, not tabs.
 
 For example:
 
+:+1:
 ```systemverilog
 logic [7:0]  my_interface_data;
 logic [15:0] my_interface_address;
 logic        my_interface_enable;
+
+logic       another_signal;
+logic [7:0] something_else;
+```
+
+:+1:
+```systemverilog
+mod u_mod (
+  .clk_i,
+  .rst_ni,
+  .sig_i          (my_signal_in),
+  .sig2_i         (my_signal_out),
+  // comment with no blank line maintains the block
+  .in_same_block_i(my_signal_in),
+  .sig3_i         (something),
+
+  .in_another_block_i(my_signal_in),
+  .sig4_i            (something)
+);
 ```
 
 #### Expressions
@@ -933,8 +959,8 @@ module my_module #(
     .clk_i,
     .rst_ni,
     .req_valid_i,
-    .req_data_i    (req_data_masked),
-    .req_ready_o,
+    .req_data_i (req_data_masked),
+    .req_ready_o(req_ready),
     ...
   );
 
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):

msfschaffner (approved):

VerilogCodingStyle.md
```diff
@@ -1527,6 +1526,33 @@ Do not use positional arguments to connect signals to ports.
...
+style. An exception is if there is only one parameter that is obvious such as
+register width, then the instantiation can be implicit.
+
+Indentation for module instantiation follows the standard indentation
+rule of two space indentation.
+
+```systemverilog
+my_module #(
+  .Height(5),
+  .Width(10)
```
sriyerg, line 1543: We should also tabular-align parameter set.

VerilogCodingStyle.md
```diff
@@ -1527,6 +1526,33 @@ Do not use positional arguments to connect signals to ports.
...
+) my_module (
+  ...etc...
+
+my_reg #(16) my_reg0 (.clk_i, .rst_ni, .d_i(data_in), .q_o(data_out));
```
sriyerg, line 1547: Verible will format this to have each port connection on a new line (which I actually like, since it makes to port connections easier to read).

VerilogCodingStyle.md
```diff
@@ -523,14 +528,33 @@ that the identical parts are directly above one another. This alignment makes it
...
+```systemverilog
+mod u_mod (
+  .clk_i,
+  .rst_ni,
+  .sig_i (my_signal_in),
+  .sig2_i(my_signal_out),
+
+  .in_another_block_i(my_signal_in),
+  .sig3_i            (something)
+);
```
mdhayter, line 557: Would it be worth an additional example with a comment rather than a blank line?

```
mod u_mod (
  .clk_i,
  .rst_ni,
  .sig_i          (my_signal_in),
  .sig2_i         (my_signal_out),
  // comment with no blank line maintains the block
  .in_same_block_i(my_signal_in),
  .sig3_i         (something)
);

File: VerilogCodingStyle.md [hunk 4]
```diff
@@ -1527,6 +1526,33 @@ Do not use positional arguments to connect signals to ports.
 
 Instantiate ports in the same order as they are defined in the module.
 
+Align expressions in [tabular style](#tabular-alignment).
```
mdhayter, line 1529: Maybe "port expressions" rather than just "expressions"? I don't have a strong view.

VerilogCodingStyle.md
imphil: Yes, but let's handle that in a separate PR. I'll push that as soon as we have merged this one to avoid the merge conflict.

VerilogCodingStyle.md
imphil: The SystemVerilog spec calls it "expression", but I also find it too generic. I updated the text to read "port expressions", as suggested.

imphil:
Thanks for your feedback @sriyerg and @mdhayter, I updated the PR to include your suggestions.

GregAC:
Do we want to explicitly require tabular alignment, and if so specify a particular style for doing so? As is this just encourages aligning ports but doesn't mandate it and leave it up to the code author to choose a style for doing it.

A strict 'you must do it like this' is problematic because we'll have lots of RTL violating the style guide, but maybe a 'ports must be aligned, style can vary but must be kept consistent across a block and our preferred style(s) is/are ...'?

imphil:
> Do we want to explicitly require tabular alignment, and if so specify a particular style for doing so?

This change is intended to require tabular alignment, which is further described in the "Tabular Alignment" section at the top of the document.

> A strict 'you must do it like this' is problematic because we'll have lots of RTL violating the style guide, but maybe a 'ports must be aligned, style can vary but must be kept consistent across a block and our preferred style(s) is/are ...'?

The intend of this clarification is to mandate one way of formatting port expressions, which can then be enforced by tooling.

I do not plan to go around and change all violating RTL code right now, we can do that as we stumble across it, or just rely on Verible format or another tool to do it for us once it is ready. (Which is the same approach we used when introducing clang-format.)

GregAC:
> This change is intended to require tabular alignment, which is further described in the "Tabular Alignment" section at the top of the document.

In that case I think we need stronger wording, the "Tabular Alignment" section starts

> Adding whitespace to cause related things to align is encouraged.

Should we switch this to

> Tabular alignment of ports of module instantiations is required.
> Adding whitespace to cause related things to align elsewhere is encouraged.

With some explanation of the desired style

> There must be no trailing or leading whitespace inside the parenthesis containing a port expression. Within a aligned port block there must be no whitespace between the longest port name and the parenthesis.

Then a couple of bad examples to demonstrate violations:

```systemverilog
mod u_mod(
  .clk_i,
  .rst_ni,

// Not allowed, leading/trailing whitespace in parenthesis
  .sig_1_i( sig_1 ),
  .sig_2_i( sig_2 )
);
```


```systemverilog
mod u_mod(
  .clk_i,
  .rst_ni,

  .short_sig_i                       (sig_1),
  // Not allowed, there should be no whitespace between the longest signal name and the parenthesis
  .a_very_long_signal_name_indeed_i  (sig_2)
);
```

imphil:
@GregAC good points. I clarified these points, PTAL.

GregAC (approved):
Looks good, thanks

imphil:
The discussion in https://github.com/lowRISC/style-guides/issues/48 has concluded and we do have a large majority for this change. Let's get it in!

PR #54, opened 2021-05-21
Exemptions on 'spaces around keywords' rule

sriyerg:
The `wait(foo);` is a complete statement in itself and should not be
following this rule. Instead it is better aligned with the 'functions
and task calls' rule.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -766,6 +766,20 @@ always_ff @(posedge clk) begin
 end
 ```
 
+However, keywords that behave more like a function or a task are better aligned
+with the [Function And Task Calls](#function-and-task-calls) rule. An example
+of this is the `wait` statement used in verification:
+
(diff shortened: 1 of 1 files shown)


VerilogCodingStyle.md
```diff
@@ -766,6 +766,20 @@ always_ff @(posedge clk) begin
 end
 ```
 
+However, keywords that behave more like a function or a task are better aligned
```
matutem, line 769: I frankly don't agree: the guideline is simple enough, and it is best to keep it simple. Notice you are making this even less clear by saying "wait" is an example, as if more exceptions can also be considered.

matutem (changes requested):
I don't think it is a good idea to take simplicity away from these rules.

sriyerg:
Let's take a quick vote on this and go with the majority team's recommendation. I filed #55 to solidify this.

PR #57, opened 2022-01-21
Discourage the use of hierarchical references in synthesiable RTL code

vogelpi:
I thought we should put this in the style guide based on the observations made in lowRISC/OpenTitan#10180.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -106,6 +106,7 @@ representation of this style guide.
     - [Functions and Tasks](#functions-and-tasks)
     - [Problematic Language Features and Constructs](#problematic-language-features-and-constructs)
       - [Floating begin-end blocks](#floating-begin-end-blocks)
+      - [Hierarchical references](#hierarchical-references)
   - [Design Conventions](#design-conventions)
     - [Summary](#summary-3)
     - [Declare all signals](#declare-all-signals)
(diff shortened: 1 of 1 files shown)


VerilogCodingStyle.md
```diff
@@ -2469,6 +2470,46 @@ module foo (
 endmodule
 ```
 
+#### Hierarchical references
+
+The use of hierarchical references in synthesizable RTL code is discouraged.
+Certain synthesis tools indeed supports this, while some tools error out
```
rswarbrick, line 2476: Nit: Should be "support"

VerilogCodingStyle.md
```diff
@@ -2469,6 +2470,46 @@ module foo (
 endmodule
 ```
 
+#### Hierarchical references
+
+The use of hierarchical references in synthesizable RTL code is discouraged.
```
rswarbrick, line 2475: We could even prohibit them. Do we have any examples where they're actually needed? If not all of our synthesis tooling supports them, I assume not?

VerilogCodingStyle.md
vogelpi: Done, thanks!

VerilogCodingStyle.md
GregAC: +1 to prohibit for synthesizable RTL.

We should definitely still allow them for assertions and such, these may be included in synthesizable RTL files but will be guarded by macros so they're stripped out when actually synthesizing.

vogelpi: IIRC we have quite a few hierarchical references in the Ibex code base. However, they are just used to implement non-synthesizable features like RVFI and the trace module.

vogelpi: Actually, also the RVFI stuff in Ibex is guarded by macros. So it would be fine to prohibit them with the exemption proposed by Greg. Let me change the PR accordingly.

vogelpi: Alright, I've changed the wording accordingly, PTAL.

GregAC (approved):

rswarbrick (approved):

msfschaffner:
Thanks for taking care of this!
I fully agree that this must not be used in synthesizable RTL code.

PR #58, opened 2022-06-10
Add guidelines for chip-level test forcing and probing

weicaiyang:
Refer to https://github.com/lowRISC/opentitan/issues/12610
Signed-off-by: Weicai Yang <weicai@google.com>


matutem (approved):
This agrees with what we discussed.

LGTM

Issue #52, opened 2021-05-04
Use of SystemVerilog Interfaces


MikeOpenHWGroup:
This guide lists SV interfaces as [problematic](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#problematic-language-features-and-constructs) and that their use is discouraged.  Do you have a specific rationale for this?

My own experience is that SV Interfaces are invaluable for verification code, and _can_ be used for RTL as long as care is taken with synthesis.  I do not know if SV Interfaces are supported by all FPGA synthesis tools (its been a while since I've tried that).

imphil:
The style guide at https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md is primarily focused on synthesizeable SV, which passes through a much larger set of tools than DV code, leading to a rather conservative choice of allowed features. We (me personally and others on OpenTitan) had various cases in the past where interfaces were mis-synthesized, partially supported, etc. 

Unfortunately I don't have specific test cases or specific bugs at hand to be able to check if that is still a valid concern, making it effectively a risk trade-off: do interfaces provide enough value to try again? For OpenTitan, we didn't find a good reason to use them again, but we'd be very interested in the experience others have!

@sjgitty @tjaychen @eunchan do you remember more details of when you last tried SV interfaces? (I think  Scott had a story there about a misinterpretation of some code by a tool very late in the design flow, leading to a hard-to-diagnose bug shortly before tapeout.)

*For DV code*, SV interfaces are usable and widely used within OpenTitan and Ibex DV, even though we don't explicitly allow them in the DV style guide at https://github.com/lowRISC/style-guides/blob/master/DVCodingStyle.md#interfaces-clocking-blocks-modports. 

@sriyerg that's something we should probably spell out more strongly?

sriyerg:
We already maximize code-reuse and reduce potential connectivity mistakes by bundling common port signals into structs. I would assume the historic decision to prefer structs over interfaces was to ensure support for a wide range of tools, including some of the open source ones. Perhaps someone can comment on whether there were tools that did not play nice with interfaces.

[...]

PR #60, opened 2022-07-20
add a reference to sv-tests

josuah:
This is a request for comments.

In #52 were discussed the use of SystemVerilog interfaces, and the reason stated was the absence of a complete tooling support. It is true that a few major tools part of the open source toolchain do not support them fully, like [yosys](https://chipsalliance.github.io/sv-tests-results/?v=yosys+25.3+interface) for which it is WIP.

What about a referencne to sv-tests? This would hint that the motivation is not that interfaces are a confusing feature, but rather a poorly supported one.

If not interesting to have it there, feel free to close this.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -2547,6 +2547,11 @@ unless otherwise noted:
 -   Interfaces.
 -   The `alias` statement.
 
+To make sure that a language feature is widely supported, tools like
+[sv-tests][svt] can be used for a non-exhaustive overview.
+
+[svt]: https://chipsalliance.github.io/sv-tests-results/
+
 #### Floating begin-end blocks
 
 The use of generate blocks other than `for` loop, `if`, or `case` generate


GregAC:
Thanks for the contribution @josuah 

sv-tests is indeed a useful tool, though it does only cover open source tools. We want blocks written using the lowRISC style guide to be broadly useable across the whole EDA ecosystem which includes many proprietary tools.

Often limitations in language usage are born out from past experience with proprietary tools. So it's not as simple as looking at a feature matrix and seeing FooSynth doesn't support interfaces (if you can even find such a feature matrix). Unfortunately due to licensing and confidentiality agreements publicly documenting specific issues with specific tools is also fraught with difficulty.

So sv-tests alone isn't sufficient for our purposes, nor could we put together a similar matrix for proprietary tools (or at least make a public one).

Though the point is taken we should provide further justification for the exclusion of interfaces. Some wording to explain that interfaces have been observed to be poorly supported in some EDA flows could suffice.

josuah:
I just noticed this remained open and I did not answer.
Thank you for the explanations once again. This makes complete sense, 
It feels good to have a frame of reference collaborated with many people using many different tools.
This actually brings visibility and transparency where it was supposed impossible to have any (proprietary toolchains with NDAs).
Acheivement!

PR #66, opened 2023-07-30
Functions should avoid non-local references

sifferman:
Hello. I'm a huge fan of this style guide, and I enforce it for all my students. Thank you!

In an Icarus issue (steveicarus/iverilog#983), we agreed that using continuous assignment from functions is inconsistent among tools and may deserve its own warning. I don't believe that the lowRISC style guide mentions this issue, so I wrote up a section on it.

### DIFF
--- a/VerilogCodingStyle.md
+++ b/VerilogCodingStyle.md
@@ -2539,6 +2539,52 @@ function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
 endfunction
 ```
 
+Functions should not reference any non-local signals or variables outside their
+scope. Avoiding non-local references improves readability and helps reduce
+simulation/synthesis mismatches. Accessing non-local parameters and constants
+is allowed.
+
+&#x1f44e;
+```systemverilog {.bad}
+// - Incorrect because `mem` is not local to get_mem()
+// - Incorrect because `in_i` is not local to get_mem()
+module mymod (
+  input   logic [7:0] in_i,
+  output  logic [7:0] out_o
+);
+
+logic [7:0] mem[256];
+
+function automatic logic [7:0] get_mem();
+  return mem[in_i];
+endfunction
+
+assign out_o = get_mem();
+
+endmodule
+```
+
+&#x1f44d;
+```systemverilog {.good}
+// - Correct because `MagicValue` is a parameter
+// - Correct because `my_pkg::OtherMagicValue` is a parameter
+// - Correct because `in_i` passed as an argument
+module mymod (
+  input   logic [7:0] in_i,
+  output  logic [7:0] out_o
+);
+
+localparam [7:0] MagicValue = 1;
+
+function automatic logic is_magic(logic [7:0] v);
+  return (v == MagicValue) || (v == my_pkg::OtherMagicValue);
+endfunction
+
+assign out_o = is_magic(in_i);
+
+endmodule
+```
+
 ### Problematic Language Features and Constructs
 
 These language features are considered problematic and their use is discouraged


VerilogCodingStyle.md
```diff
@@ -2610,6 +2611,79 @@ module mymod (
...
+
+&#x1f44d;
+```systemverilog {.good}
+module mymod (
+  input        in_i,
+  output logic out_o
+);
+
+function automatic myfunc;
+  return in_i;
```
nbdd0121, line 2659: I think the fact that functions reference anything non-local is bad.

The whole "continuous assignment from function" issue is moot if we just disallow the function from referencing anything non-local.

sifferman:
Thanks for the feedback! Yes, it seems that openTitan, Ibex, and CVA6 already follow that stricter rule.

I did find one exception in Ibex, but it is for DV, so it probably doesn't matter:
https://github.com/lowRISC/ibex/blob/f60d03b6b064c38e111931ff9b0bae70d6af9af1/examples/simple_system/rtl/ibex_simple_system.sv#L323-L325

sifferman:
How does this look?

VerilogCodingStyle.md
```diff
@@ -2539,6 +2539,52 @@ function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
 endfunction
 ```
 
+Functions should not reference any non-local signals, variables, or memories
```
nbdd0121, line 2542: I would leave out "memories" because it's really just a variable that is an unpacked array.

VerilogCodingStyle.md
```diff
@@ -2539,6 +2539,52 @@ function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
...
+outside their scope. Avoiding non-local references improves readability and
+helps reduce simulation/synthesis mismatches. Accessing non-local parameters
```
nbdd0121, line 2544: Maybe add "and constants" because they are of similar nature (as in, would not change at runtime).

VerilogCodingStyle.md
```diff
@@ -2539,6 +2539,52 @@ function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
...
+&#x1f44e;
+```systemverilog {.bad}
+// - Incorrect because `MEM` is not local to get_mem()
+// - Incorrect because `in_i` is not local to get_mem()
+module mymod (
+  input   logic [7:0] in_i,
+  output  logic [7:0] out_o
+);
+
+logic [7:0] MEM[256];
```
nbdd0121, line 2556: Note that this is against the coding style (it should be lower snake case)

GregAC:
> I did find one exception in Ibex, but it is for DV, so it probably doesn't matter:
https://github.com/lowRISC/ibex/blob/f60d03b6b064c38e111931ff9b0bae70d6af9af1/examples/simple_system/rtl/ibex_simple_system.sv#L323-L325

Yup that's fine as it's not used by synthesisable code. It exists to give the testbench an easy way to read out performance counters.

nbdd0121 (approved):

VerilogCodingStyle.md
```diff
@@ -2539,6 +2539,52 @@ function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
...
+// - Correct because `my_pkg::OtherMagicValue` is a parameter
+// - Correct because `in_i` passed as an argument
+module mymod (
+  input   logic [7:0] in_i,
+  output  logic [7:0] out_o
+);
+
+localparam [31:0] MagicValue = 1;
+
+function automatic logic is_magic(logic [31:0] v);
```
marnovandermaas, line 2579:
```suggestion
function automatic logic is_magic(logic [7:0] v);
```
Minor comment: maybe match this with the bit width of `in_i`?

VerilogCodingStyle.md
```diff
@@ -2539,6 +2539,52 @@ function automatic logic [2:0] foo(logic [2:0] a, logic [2:0] b);
...
+```systemverilog {.good}
+// - Correct because `MagicValue` is a parameter
+// - Correct because `my_pkg::OtherMagicValue` is a parameter
+// - Correct because `in_i` passed as an argument
+module mymod (
+  input   logic [7:0] in_i,
+  output  logic [7:0] out_o
+);
+
+localparam [31:0] MagicValue = 1;
```
marnovandermaas, line 2577:
```suggestion
localparam [7:0] MagicValue = 1;
```
Minor comment: maybe match the bit width with `in_i`?

marnovandermaas (commented):
Thanks for making these changes, I think this is a really valuable addition to the style guide. I just added a few minor comments. Would you mind squashing your commits into one commit? If you don't know how to do that, I can do it for you.

sifferman:
Just squashed. Thanks for all the feedback!

Although I wanted to get your opinion on another issue with functions. VCS will currently not run this correctly:

```systemverilog
module test;

function automatic [7:0] test_func;
    test_func = 1;
endfunction

wire [7:0] test_wire = test_func();

initial begin
    #1;
    $display("Expected %h, Recieved %h", test_func(), test_wire);
    // Will fail because test_func() is never run
    $finish;
end

endmodule
```

Maybe this should be a separate issue, but I wonder if there should be another rule against blanket calling of functions from assign/wire. It works fine if `always_comb` is used.

rswarbrick (approved):
This looks excellent! Thank you very much for the improvement.

marnovandermaas (approved):
Thanks for making these changes!

rswarbrick:
Following up on the question about functions and continuous assignments: I didn't know that VCS wasn't happy with that! I just looked at IEEE 1800, and there's no mention of a constraint: the right hand side of a "net declaration assignment" seems to be an arbitrary expression in the syntax, at least.

If we want to think carefully about style guide recommendations for this, maybe we should start a separate issue.

marnovandermaas:
> Maybe this should be a separate issue, but I wonder if there should be another rule against blanket calling of functions from assign/wire. It works fine if `always_comb` is used.

We can probably add this in a separate PR, but it sounds like we should restrict assigning a functions output directly to a wire and mandate that this happens in an always_comb block.

GregAC:
I think I'd prefer to outright ban `wire [7:0] test = test_func()` type statements. This is because their semantics varies depending upon the net type. For `wire` you do get the equivilent to a continuous assignment but for `logic` it's just an initial value setting (much like an `initial` block though not identical! Those initial assignments are made before any `initial` blocks are run). Switching `wire` to `logic` may otherwise appear innocuous so better to avoid the complexity altogether. I note we explicitly allowed this currently: 

https://github.com/lowRISC/style-guides/blob/5cd912c617c712c773fdc9da9259298d183a2da9/VerilogCodingStyle.md?plain=1#L2702-L2725

Initial assignments can be useful in FPGA but are deadly for ASIC synthesis as it's a great way to get simulation/synthesis mismatch where the FPGA version agrees with the simulation but the ASIC synthesis doesn't. If you confine them to an `'initial` block at least it makes it far obvious you have one. We could consider saying `initial` blocks are only allowed in RTL that is strictly for FPGA usage (e.g. top-levels and wrappers etc that we may use on FPGA to instantiate IP)

nbdd0121:
I personally don't think we should get rid of the wire shorthand. It is orthogonal to the reported VCS issue.

The VCS issue probably is caused by the function being zero argument (and therefore not sensitive to anything). I think we should ban functions with no input instead. It could be a localparam instead.

a-will:
> I think I'd prefer to outright ban `wire [7:0] test = test_func()` type statements. This is because their semantics varies depending upon the net type. For `wire` you do get the equivilent to a continuous assignment but for `logic` it's just an initial value setting (much like an `initial` block though not identical! Those initial assignments are made before any `initial` blocks are run). Switching `wire` to `logic` may otherwise appear innocuous so better to avoid the complexity altogether. I note we explicitly allowed this currently:

(pedantic mode on) `wire` is a "net type" and `logic` is a "data type". Two different things. This is frequently taught incorrectly in universities, and it's a pet peeve of mine.

Unfortunately, because many, many people have received the same poor instruction, that `logic` should just be used everywhere and can replace the `wire` vs. `reg` confusion, we're left with a lot of bad practices and more confusion. The differences between nets and variables are still relevant.

This language maybe gets a little too cute with defaults:

```systemverilog
// Equivalent variable declarations with initial values.
logic foo = 1'b0;
var logic foo = 1'b0;

// Equivalent net declarations with continuous assignment.
wire bar = 1'b0;
wire logic bar = 1'b0;
```

sifferman:
I just created a new issue for this. We can discuss more here: #67

PR #74, opened 2024-10-31
Remove `_i` and `_o` naming convention

mgottscho:


mgottscho:
My bad, meant to open this PR only on my own fork!

ballifatih:
As far as I understand the removal of `_i` of `_o` is for your own use, so it doesn't really change lowrisc's own coding guide. So my comments on this closed PR is only some casual discussion points.

I also self debated whether adding `_i` and `_o` to naming convention clutters module definitions. I always thought in the past that they made the code less cleaner so it would be an improvement to remove them.

It was only when I started reviewing a large volume of RTL code that I realized how important and useful these `_i` and `_o` suffixes are. As I navigate through multiple top-level modules, directly being able to see whether a signal is an input or output is a non-trivial speed-up.

I would also advocate for keeping `_i` and `_o` convention elsewhere and I just wanted to share my two cents.

PR #88, opened 2025-06-05
DV style guide: Do not use illegal_bins as a checker

antmarzam:


DVCodingStyle.md
```diff
@@ -1180,6 +1181,13 @@ backdoor force or probe. This assumes gate-level netlist won't be flattened.
   clocks/resets are in the module inputs/outputs. If we have to, place it in an anchored buffer.
 * CSR hierarchies are likely to be preserved, so CSR backdoor access will still work.
 
+### Functional coverage
+
+1. Do not use `illegal_bins` as a checker. If the condition being illegal should cause a simulation
```
marnovandermaas, line 1186: No need to number this if there is only one thing.

DVCodingStyle.md
```diff
@@ -1180,6 +1181,13 @@ backdoor force or probe. This assumes gate-level netlist won't be flattened.
...
+   failure, ensure there is also an associated check with it. It's OK to use `illegal_bins` as the default case,
```
marnovandermaas, line 1187: Maybe give an example of what that check could be like an assertion.

DVCodingStyle.md
```diff
@@ -1180,6 +1181,13 @@ backdoor force or probe. This assumes gate-level netlist won't be flattened.
...
+   but not to rely in it as a check since it will only be active when coverage is enabled. In addition,
```
marnovandermaas, line 1188: in -> on

marnovandermaas (commented):
A few comments from my end. Can you also link the issue where this was discussed?

DVCodingStyle.md
antmarzam: Removed the number, thank you!

marnovandermaas:
Linked to the following issue: https://github.com/lowRISC/style-guides/issues/87
What's our policy on changing these guidelines when OpenTitan doesn't comply with them? Should we first fix the OpenTitan codebase?

rswarbrick:
I think I'd probably go for "solution 1"

martin-velay:
Yes, I don't mind so much of the way, as long as we're sure the issue will be raised properly

antmarzam:
I just pushed the changes clarifying a bit where the check could be carried out. Please have a look and see if you're happy with it :)

martin-velay (approved):
LGTM, thanks Antonio!

rswarbrick:
Hi Antonio! What's the status of this PR? I know we chatted quite a bit about this sort of thing over the summer. Is there a reason we didn't go ahead and merge the change?

PR #91, opened 2026-04-21
[dv] Improve description of how to use fork/join

rswarbrick:
- Give a proper example for how to give an isolation fork.

- Avoid giving references for why not to use "disable fork_label" (this document is instruction, not persuasion!)

- Avoid using "thread". The noun *is* used in IEEE 1800, but I don't think it's defined anywhere and I think it is being used as a synonym for "process".

- Give an explicit example of how to avoid the syntax for nested forks getting too enormous.

- Add instruction about not following a begin/end block with a bare item.


rswarbrick:
@vogelpi: After our messages last week, here's a proposed guideline on how to structure fork/join blocks properly (after a discussion in the DV team meeting last Friday).

DVCodingStyle.md
```diff
@@ -1073,32 +1073,83 @@ assert(function_that_has_side_effects());
...
+join
 ```
-fork: fork_label
-  begin: thread_label
+
+The isolation fork ensures that the `disable fork` kills any processes
+inside the fork that are still running, but does not kill unrelated
+processes that happen to be siblings of the one that runs the example.
+
+Use `disable fork` (inside an isolation fork) instead of disabling an
```
martin-velay: Very nit:

```suggestion
Use `disable fork` (inside an isolation fork) instead of disabling a
```

martin-velay (approved):
Thanks for making this clearer. Everything LGTM

DVCodingStyle.md
rswarbrick: Gaargh! :-)

PR #93, opened 2026-06-26
SV style guide: Add missing `end` keywords

glaserf:


rswarbrick (approved):
These are quite amusing (and go back to the dawn of time in 2019!) :-)

Looks good to me.

